PR #120, opened 2024-06-30
docs: add layered settings docs & guides

obs-gh-kylechamplin:
Request from CBA to document the layered settings for a few use cases like setting query governor limits, transform governor limits, and datastream retention.


docs/resources/layered_setting_record.md
```diff
@@ -9,6 +9,221 @@ description: |-
...
+      "default_value": "0.0",
+      "possible_target_scopes": ["Customer"],
+      "setting": "TransformGovernor.creditsPerDay",
+      "type": "float64",
+      "writable_by": ["admin", "superadmin"]
+    }
+
+```
+
+## Example Usage
```
vikramraman, line 91: terraform docs includes an option to specify guides for such detailed documentation. For ex: see [here](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/guides/monitors) 

It would be preferable to retain one example usage here, and port the rest under guides.
To do so:
- we'll need to add a relevant template file under `templates/guides` directory (for ex: [here](https://github.com/DataDog/terraform-provider-datadog/tree/master/templates/guides))
- And relevant terraform snippets that are referenced from the template file under `examples/guides` (for ex: [here](https://github.com/DataDog/terraform-provider-datadog/tree/master/examples/guides))

vikramraman (changes requested):

obs-gh-kylechamplin (commented):
RGR - can we pursue this approach a bit later as I do think its worthwhile. Right now the layered settings "documentation" is not sufficient for customers to actually use it as is, I think this is a quick but effective bandaid while we look at moving toward those guides and examples.

vikramraman (approved):

PR #121, opened 2024-07-01
Monitorv2 resource

obs-gh-owengoebel:


observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,962 @@
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{ // MonitorV2RuleInput!
+						"level": { // MonitorV2AlarmLevel!
+							//   Critical
+							//   Error
+							//   Informational
+							//   None
+							//   Warning
+							Type:     schema.TypeString,
+							Required: true,
```
vikramraman, line 111: Ditto, add validation for this field

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,962 @@
...
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{ // MonitorV2ComparisonInput
+												"compare_fn": { // MonitorV2ComparisonFunction!
+													//   Equal
+													//   Greater
+													//   GreaterOrEqual
+													//   Less
+													//   LessOrEqual
+													//   NotEqual
+													Type:     schema.TypeString,
```
vikramraman, line 133: Same here: validation

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,962 @@
...
+										MinItems: 1,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"compare_fn": { // MonitorV2ComparisonFunction!
+													//   Equal
+													//   Greater
+													//   GreaterOrEqual
+													//   Less
+													//   LessOrEqual
+													//   NotEqual
```
vikramraman, line 192: validation

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,962 @@
...
+													MinItems: 1,
+													Elem: &schema.Resource{
+														Schema: map[string]*schema.Schema{ // MonitorV2ComparisonInput!
+															"compare_fn": { // MonitorV2ComparisonFunction!
+																//   Equal
+																//   Greater
+																//   GreaterOrEqual
+																//   Less
+																//   LessOrEqual
+																//   NotEqual
```
vikramraman, line 270: validation

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,962 @@
...
+				},
+			},
+			"scheduling": { // MonitorV2SchedulingInput
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				MinItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"interval": { // MonitorV2IntervalScheduleInput
```
vikramraman, line 483: Is `scheduling` supposed to contain either `interval` or `transform` or it can have both?

If either, you can specify the `ExactlyOneOf` property for both interval and transform. There are other places where this is used that you can grep for reference.

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,962 @@
...
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceMonitorV2Create(ctx context.Context, data *schema.ResourceData, meta interface{}) (diags diag.Diagnostics) {
+	client := meta.(*observe.Client)
+
+	fmt.Printf("\n\n\n\n\nstart of create\n\n\n\n\n")
```
vikramraman, line 544: Remove all the debug statements once you're done validating

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,962 @@
...
+func resourceMonitorV2Update(ctx context.Context, data *schema.ResourceData, meta interface{}) (diags diag.Diagnostics) {
+	client := meta.(*observe.Client)
+
+	input, diags := newMonitorV2Input(data)
+	if diags.HasError() {
+		return diags
+	}
+
+	_, err := client.UpdateMonitorV2(ctx, data.Id(), input)
+	if err != nil {
```
vikramraman, line 573: We will want to handle the not found case similar to how we added for attachments.

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,962 @@
...
+		return diag.Errorf("failed to create monitor: %s", err.Error())
+	}
+
+	return append(diags, resourceMonitorV2Read(ctx, data, meta)...)
+}
+
+func resourceMonitorV2Read(ctx context.Context, data *schema.ResourceData, meta interface{}) (diags diag.Diagnostics) {
+	client := meta.(*observe.Client)
+
+	monitor, err := client.GetMonitorV2(ctx, data.Id())
```
vikramraman, line 583: Ditto: Need to handle the not found case. Assuming the api changes you made to gmodelgen should have automatically added it for monv2? Even otherwise, it's good to add the check here on the client side so we are handling it from the beginning.

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,1036 @@
...
+													//   LessOrEqual
+													//   NotEqual
+													Type:     schema.TypeString,
+													Required: true,
+												},
+												"type": {
+													Type:     schema.TypeString,
+													Required: true,
+												},
+												"value_int64": { // Int64
```
vikramraman, line 140: Can likely specify ExactlyOneOf for all these value fields if only one of them should be specified based on the type

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,1036 @@
...
+																//   NotEqual
+																Type:     schema.TypeString,
+																Required: true,
+															},
+															"type": {
+																Type:     schema.TypeString,
+																Required: true,
+															},
+															"value_int64": { // Int64
+																Type:     schema.TypeInt,
```
vikramraman, line 279: Similar comment to explore ExactlyOneOf for all the value fields

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,1036 @@
...
+
+	if err := data.Set("description", monitor.Description); err != nil {
+		diags = append(diags, diag.FromErr(err)...)
+	}
+
+	if err := data.Set("id", monitor.Oid().String()); err != nil {
+		diags = append(diags, diag.FromErr(err)...)
+	}
+
+	return diags
```
vikramraman, line 609: Just clarifying you'll be including the other fields on the monitor?

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,1036 @@
...
+		kinds = append(kinds, "value_duration")
+	}
+	if typeName == "value_timestamp" && hasTimestamp && valueTimestamp != nil {
+		tsp, _ := time.Parse(time.RFC3339, valueTimestamp.(string))
+		tss := types.TimeScalar(tsp)
+		ret.Timestamp = &tss
+		nvalue++
+		kinds = append(kinds, "value_timestamp")
+	}
+	if nvalue == 0 {
```
vikramraman, line 1029: As commented above where these fields are defined, you can rely on ExactlyOneOf on the fields themselves and let the provider validate that at least one of these is provided. 

So you can simplify this function to just read / return the relevant value.

vikramraman (changes requested):
Looks good overall! Left some early comments.

observe/resource_monitor_v2.go
obs-gh-owengoebel: yes. Just stubbed it out for now, but I currently am working on this

observe/resource_monitor_v2.go
obs-gh-owengoebel: Will look into this

observe/descriptions/monitorv2.yaml
```diff
@@ -0,0 +1,44 @@
+# description:
+#   Monitors provide a flexible way to alert for patterns in your
```
obs-gh-catherman, line 2: This comment is probably copy/pasta'd from v1, but doesn't really reflect v2 accurately. Can suggestion new text...

```
Monitors provide a configurable way to alert when conditions about incoming data
are matched. These alerts can optionally also be forwarded to notification receivers
like email and webhooks using shared or single-monitor actions to configure the
template and destinations to configure the receiver.
```

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,1076 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	observe "github.com/observeinc/terraform-provider-observe/client"
+	"github.com/observeinc/terraform-provider-observe/client/meta"
+	gql "github.com/observeinc/terraform-provider-observe/client/meta"
+	"github.com/observeinc/terraform-provider-observe/client/meta/types"
+	"github.com/observeinc/terraform-provider-observe/client/oid"
+)
+
+// TODO: make the schema keys constants?
```
obs-gh-catherman, line 18: is this comment necessary?

observe/resource_monitor_v2.go
obs-gh-catherman: Exactly one is correct.

obs-gh-catherman: Let's make `scheduling` not optional. It's optional in the API, but let's make people explicitly set it in TF. (Also, maybe this is a TF thing, but MinItems: 1 but optional is weird to me :) )

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,1076 @@
...
+
+func resourceMonitorV2Create(ctx context.Context, data *schema.ResourceData, meta interface{}) (diags diag.Diagnostics) {
+	client := meta.(*observe.Client)
+
+	input, diags := newMonitorV2Input(data)
+	if diags.HasError() {
+		return diags
+	}
+
+	id, _ := oid.NewOID(data.Get("workspace_id").(string))
```
obs-gh-catherman, line 579: Should we just add a `MustNewOID` and panic if it fails for these cases where we expect it to always work?

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,1076 @@
...
+func newMonitorV2PrimitiveValue(path string, data *schema.ResourceData, ret *gql.PrimitiveValueInput) diag.Diagnostics {
+
+	valueBool, hasBool := data.GetOk(fmt.Sprintf("%svalue_bool", path))
+	valueInt, hasInt := data.GetOk(fmt.Sprintf("%svalue_int64", path))
+	valueFloat, hasFloat := data.GetOk(fmt.Sprintf("%svalue_float64", path))
+	valueString, hasString := data.GetOk(fmt.Sprintf("%svalue_string", path))
+	valueDuration, hasDuration := data.GetOk(fmt.Sprintf("%svalue_duration", path))
+	valueTimestamp, hasTimestamp := data.GetOk(fmt.Sprintf("%svalue_timestamp", path))
+
+	//	NOTE: I rely on the fact that sizeof(int) == sizeof(int64) on modern systems
```
obs-gh-catherman, line 1028: did you copy/paste this from somewhere? I'm not sure what it pertains to :)

obs-gh-owengoebel: Yeah, I initially just called `primitiveValueEncode` instead of making `newMonitorV2PrimitiveValue` its own function, but then I found I wanted enough differences (ability to pass 0 for an integer, ability to specify a path) that I copied the function and made my changes. This comment is residue from that and I don't really know what the person who wrote it intended by it.

observe/resource_monitor_v2.go
obs-gh-owengoebel: I wrote it in the mindset of "you don't have to provide a list, but if you do, then it must have at least one element" because I felt like opting not to provide a list is not the same thing as opting to provide a list with 0 items. But I read that TF ignores minItems if optional is set, so I will remove minItems on all such lists

observe/resource_monitor_v2.go
obs-gh-owengoebel: unfortunately cannot do this for entries in a variable-length list because ExactlyOneOf requires an exact path

observe/resource_monitor_v2_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+// todo write some tests
+// starting point: count, threshold, ????????????
+
+var monitorV2ConfigPreamble = configPreamble + datastreamConfigPreamble
+
+func TestAccObserveMonitorV2(t *testing.T) {
```
vikramraman, line 16: Could we add a few more tests to cover the other kinds of monitors: threshold and promote?

observe/resource_monitor_v2.go
```diff
@@ -0,0 +1,1165 @@
...
+			},
+			"icon_url": { // String
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"description": { // String
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"managed_by_id": { // ObjectId
```
vikramraman, line 57: Remove this. This is handled globally at the provider level via an environment variable. 
And if provided, this is set under the `client/createMonitorV2()` function. Refer to how we do this for monitorv1 for example.

client/api.go
```diff
@@ -494,6 +494,40 @@ func (c *Client) GetMonitor(ctx context.Context, id string) (*meta.Monitor, erro
...
+	}
+	return c.Meta.CreateMonitorV2(ctx, workspaceId, input)
+}
+
+func (c *Client) UpdateMonitorV2(ctx context.Context, id string, input *meta.MonitorV2Input) (*meta.MonitorV2, error) {
+	if !c.Flags[flagObs2110] {
+		c.obs2110.Lock()
+		defer c.obs2110.Unlock()
+	}
+	if c.Config.ManagingObjectID != nil {
```
vikramraman, line 513: We don't usually allow mutating the managing object after creation. So this can be removed for now.

observe/descriptions/monitorv2.yaml
```diff
@@ -0,0 +1,105 @@
+description:
+  NOTE. This feature is still in development. It is not meant for customer use yet.
```
vikramraman, line 2: Could this be `NOTE:` instead of `NOTE.`. Or does that break the doc generation in some way.

vikramraman (changes requested):
Looking good! Just a few minor nits.

observe/descriptions/monitorv2.yaml
obs-gh-owengoebel: I didn't take the time to figure out how to escape a colon in a YAML string. This is an easy fix though, will push a change

vikramraman (approved):
lgtm. Thanks for pushing through all the feedback.

PR #125, opened 2024-07-15
feat: add cloudwatch metrics poller

jta:


observe/descriptions/poller.yaml
```diff
@@ -0,0 +1,51 @@
...
+      description: |
+        Query specification.
+      namespace: |
+        AWS Metric Namespace to query.
+      metric_names: |
+        Metric names to filter down to. If more than one metric name is provided, `ListMetrics` will be called with no filter on metric names.
+      dimension: |
+        Dimension filter to set. A metric must match all provided dimension filters in order to be queried for data points.
+      resource_filter:
+        description: |
```
obs-gh-alexlew, line 37: can this be marked as an array?

obs-gh-alexlew (approved):

observe/descriptions/poller.yaml
jta: This file is solely for descriptions, we retrieve the value in source code via the descriptions package, e.g. `										descriptions.Get("poller", "schema", "cloudwatch_metrics", "query", "resource_filter", "description")`.

The generated docs extract the full schema during `make generate`, whereby the provider is compiled and the schema is retrieved over RPC.

PR #126, opened 2024-07-23
feat: add monitorv2 datasource

obs-gh-owengoebel:


client/api.go
```diff
@@ -525,6 +525,10 @@ func (c *Client) GetMonitorV2(ctx context.Context, id string) (*meta.MonitorV2,
 	return c.Meta.GetMonitorV2(ctx, id)
 }
 
+func (c *Client) LookupMonitorV2(ctx context.Context, workspaceId *string, folderId *string, nameExact *string, nameSubstring *string) (*meta.MonitorV2, error) {
```
vikramraman, line 528: Even if the actual api has all these fields, we probably only want to support workspaceId and nameExact in here.

observe/data_source_monitor_v2.go
```diff
@@ -0,0 +1,451 @@
...
+
+func dataSourceMonitorV2() *schema.Resource {
+	return &schema.Resource{
+		Description: descriptions.Get("monitorv2", "description"),
+		ReadContext: dataSourceMonitorV2Read,
+		Schema: map[string]*schema.Schema{
+			// used to lookup the monitor
+			// monitor can be looked up either by providing an ID
+			// or by providing search params that can uniquely ID the monitor.
+			"get_id": { // ObjectId!
```
vikramraman, line 23: We want to be consistent with the naming across data sources, we typically have: `id`, `workspace`, `name` as the fields for looking up stuff, and should go with the same here.

observe/data_source_monitor_v2.go
```diff
@@ -0,0 +1,451 @@
...
+				Type:             schema.TypeString,
+				Optional:         true,
+				ValidateDiagFunc: validateID(),
+				Description:      descriptions.Get("common", "schema", "id"),
+			},
+			"search_workspace_id": { // ObjectId!
```
vikramraman, line 29: `workspace`

observe/data_source_monitor_v2.go
```diff
@@ -0,0 +1,451 @@
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: descriptions.Get("monitorv2", "schema", "workspace_id"),
+			},
+			"search_name_exact": {
```
vikramraman, line 34: `name`

observe/data_source_monitor_v2.go
```diff
@@ -0,0 +1,451 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			// fields of MonitorV2Input excluding the components of MonitorV2Definition
+			"workspace_id": { // ObjectId!
```
vikramraman, line 39: Can be removed, it is redundant with the `workspace` field.

observe/data_source_monitor_v2.go
```diff
@@ -0,0 +1,451 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: descriptions.Get("monitorv2", "schema", "comment"),
+			},
+			"rule_kind": { // MonitorV2RuleKind!
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: descriptions.Get("monitorv2", "schema", "rule_kind"),
+			},
+			"name": { // String!
```
vikramraman, line 54: same, redundant given `name` above

observe/data_source_monitor_v2.go
```diff
@@ -0,0 +1,451 @@
...
+										Computed:    true,
+										Description: descriptions.Get("monitorv2", "schema", "scheduling", "transform", "freshness_goal"),
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			"id": { // ObjectId!
```
vikramraman, line 260: We want to include an `oid` instead as a computed field.

vikramraman (changes requested):

docs/data-sources/monitor_v2.md
```diff
@@ -0,0 +1,411 @@
...
+template and destinations to configure the receiver.
+
+
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Optional
+
+- `data_stabilization_delay` (String) expresses the minimum time that should elapse before data is considered "good enough" to evaluate. Choosing a delay really depends on the expectations of latency of data and whether data is expected to arrive later than other data and thus would change previously evaluated results.
```
vikramraman, line 29: This looks odd, it should be computed and not be part of the optional input for this data source

docs/data-sources/monitor_v2.md
```diff
@@ -0,0 +1,411 @@
...
+- `id` (String) Resource ID for this object.
+- `lookback_time` (String) optionally describes a duration that must be satisifed by this monitor. It applies to all rules, but is only applicable to rule kinds that utilize it.
```
vikramraman, line 31: Similar comment here, this should not be needed

vikramraman (changes requested):

docs/data-sources/monitor_v2.md
obs-gh-owengoebel: makes sense. I think I forgot to delete "Optional: true" when adding "Computed: true". Fixed now

vikramraman (approved):

PR #128, opened 2024-07-24
feat: add observe_monitor_v2_action resource

obs-gh-owengoebel:


observe/resource_monitor_v2_action.go
```diff
@@ -0,0 +1,390 @@
...
+
+func resourceMonitorV2Action() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceMonitorV2Create,
+		ReadContext:   resourceMonitorV2Read,
+		UpdateContext: resourceMonitorV2Update,
+		DeleteContext: resourceMonitorV2Delete,
+		Schema: map[string]*schema.Schema{
+			// needed as input to CreateMonitorV2Action
+			"workspace_id": { // ObjectId!
```
vikramraman, line 25: just `workspace`

observe/resource_monitor_v2_action.go
```diff
@@ -0,0 +1,390 @@
...
+				Type:             schema.TypeString,
+				ForceNew:         true,
+				Required:         true,
+				ValidateDiagFunc: validateOID(oid.TypeWorkspace),
+			},
+			// fields of MonitorV2ActionInput
+			"inline": { // Boolean
```
vikramraman, line 32: I believe we only want to support shared actions. So perhaps we should not expose this, and default to false internally.

observe/resource_monitor_v2_action.go
```diff
@@ -0,0 +1,390 @@
...
+			"icon_url": { // String
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"description": { // String
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			// end of monitorV2ActionInput
+			"id": { // ObjectId!
```
vikramraman, line 66: `oid` instead of `id`

observe/resource_monitor_v2_action.go
```diff
@@ -0,0 +1,390 @@
...
+		},
+	}
+}
+
+func monitorV2EmailActionInput() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"subject": { // String
+				Type:     schema.TypeString,
+				Optional: true,
```
vikramraman, line 79: Should the `subject` and `body` be required for an email action? Feels odd that all of these are optional for an email input.

observe/resource_monitor_v2_action.go
```diff
@@ -0,0 +1,390 @@
...
+		},
+	}
+}
+
+func monitorV2WebhookActionInput() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"headers": { // [MonitorV2WebhookHeaderInput!]
+				Type:     schema.TypeList,
+				Optional: true,
```
vikramraman, line 100: Similar question here, should these be required for a webhook input?

observe/resource_monitor_v2_action.go
```diff
@@ -0,0 +1,390 @@
...
+			},
+			"type": { // MonitorV2ActionType!
+				Type:             schema.TypeString,
+				ValidateDiagFunc: validateEnums(gql.AllMonitorV2ActionTypes),
+				Required:         true,
+			},
+			"email": { // MonitorV2EmailDestinationInput
+				Type:     schema.TypeList,
+				MaxItems: 1,
```
vikramraman, line 43: Should add `ExactlyOneOf: []string{"email", "webhook"}` here and for webhook?

observe/resource_monitor_v2_action.go
```diff
@@ -0,0 +1,390 @@
...
+func resourceMonitorV2ActionDelete(ctx context.Context, data *schema.ResourceData, meta interface{}) (diags diag.Diagnostics) {
+	client := meta.(*observe.Client)
+	if err := client.DeleteMonitorV2Action(ctx, data.Id()); err != nil {
+		return diag.Errorf("failed to delete monitor action: %s", err.Error())
+	}
+	return diags
+}
+
+func resourceMonitorV2ActionRead(ctx context.Context, data *schema.ResourceData, meta interface{}) (diags diag.Diagnostics) {
+	// TODO
```
vikramraman, line 182: Remove the TODO?

vikramraman (changes requested):

observe/resource_monitor_v2_action.go
obs-gh-owengoebel: They are optional in the schema. Also in general emails don't need to have a subject and a body. I can send an email to myself without specifying either

observe/resource_monitor_v2_action.go
obs-gh-owengoebel: the schema declares these as [MonitorV2WebhookHeader!], which means the user is not required to provide any elements

obs-gh-owengoebel: Resolved. The API fills with defaults if not provided.

observe/resource_monitor_v2_action.go
obs-gh-owengoebel: Resolved. The API fills with defaults if not provided.

vikramraman (approved):

PR #130, opened 2024-07-29
docs: add cloudwatch metrics docs

jta:


docs/resources/poller.md
```diff
@@ -298,3 +298,225 @@ Import is supported using the following syntax:
 ```shell
 terraform import observe_poller.example 1414010
 ```
+
+## CloudWatch Metrics poller
```
vikramraman, line 302: This feels like it would be a better fit as a guide rather than a resource doc?

We added one for layered settings earlier for reference: https://github.com/observeinc/terraform-provider-observe/pull/120

jta: I'm not sure.. I started down that route, but it did feel like it would be too detached from the place I would expect the example to live in. If I go to the poller docs, I'd expect an example for each poller type. I'm OK either way, will let Nathan decide how best to organize it.

vikramraman: Agree, we need some pointer from the resource docs to the relevant guides to make it easier to find

PR #134, opened 2024-08-02
feat: add ability to link monitorv2 to shared actions

obs-gh-owengoebel:


observe/resource_monitor_v2.go
```diff
@@ -273,6 +270,14 @@ func resourceMonitorV2() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"actions": { // [ObjectId]
```
vikramraman, line 273: Would be good to add a description for this field

observe/resource_monitor_v2.go
```diff
@@ -456,6 +461,14 @@ func resourceMonitorV2Create(ctx context.Context, data *schema.ResourceData, met
 		return diag.Errorf("failed to create monitor: %s", err.Error())
 	}
 
+	// TODO: delete the "if" statement after the Save API call is changed to accept empty lists
```
vikramraman, line 464: I guess these can be removed now?

observe/resource_monitor_v2.go
```diff
@@ -477,8 +490,16 @@ func resourceMonitorV2Update(ctx context.Context, data *schema.ResourceData, met
 			}
 			return nil
 		}
-		return diag.Errorf("failed to create monitor: %s", err.Error())
+		return diag.Errorf("failed to update monitor: %s", err.Error())
+	}
+
+	// TODO: delete the "if" statement after the Save API call is changed to accept empty lists
```
vikramraman, line 496: Ditto, remove the TODO / commented lines?

observe/resource_monitor_v2_action_test.go
```diff
@@ -52,6 +52,7 @@ func TestAccObserveMonitorV2ActionEmail(t *testing.T) {
 								randomize = "0"
 							}
 						}
+						actions = [observe_monitor_v2_action.act.oid]
```
vikramraman, line 55: Could we add a test with multiple actions. Either email or webhook is fine. 
Whichever is easier to test for this case.

vikramraman (changes requested):

observe/resource_monitor_v2_action_test.go
```diff
@@ -183,3 +191,135 @@ func TestAccObserveMonitorV2ActionWebhook(t *testing.T) {
...
+						}
+						scheduling {
+							interval {
+								interval = "15m"
+								randomize = "0"
+							}
+						}
+						actions {
+							oid = observe_monitor_v2_action.act1.oid
+							levels = ["critical"]
```
vikramraman, line 241: Interesting, does this work? We only have an informational rule amongst the rules block.
I would think the api should error out if we refer to an unknown severity?

obs-gh-owengoebel: The test in question passed. I'm pretty sure the answer here is that the API allows triggers to be defined even if they aren't accessible via the current rules.

obs-gh-owengoebel: I have also produced this locally without issues

obs-gh-owengoebel: But for the sake of having the test make sense in mapping to a real use-case, I'll change this to informational

vikramraman (approved):

PR #141, opened 2024-08-19
fix: update monitor v2 gql schemas

vikramraman:


obs-gh-andrewlee (approved):
Looks great, thank you for the updates!

PR #147, opened 2024-08-22
fix: make monitor_v2 scheduling block optional

obs-gh-abhinavpappu:
The GraphQL currently returns null for the `scheduling` field by default (if the user hasn't modified it), resulting in the exported terraform missing the required `scheduling` block.

Talked to @obs-gh-andrewlee and it looks like this is expected, with the scheduling defaulting to `transform` if it's not provided.

### DIFF
--- a/docs/data-sources/monitor_v2.md
+++ b/docs/data-sources/monitor_v2.md
@@ -43,7 +43,7 @@ stage pipelines.
 - `oid` (String)
 - `rule_kind` (String) Describes the type of each of the rules in the definition (they must all be the same type).
 - `rules` (Block List) All rules for this monitor must be of the same MonitorRuleKind as specified in ruleKind. Rules should be constructed logically such that a state transition null->Warning implies transition from null->Informational. (see [below for nested schema](#nestedblock--rules))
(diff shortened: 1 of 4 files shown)


obs-gh-abhinavpappu:
Actually when trying to apply a terraform plan that doesn't contain a `scheduling` block, I'm seeing the error "monitor schedule is not set or is not supported" from GraphQL, which is coming from [here](https://gerrit.observeinc.com/plugins/gitiles/observe/+/refs/heads/master/code/go/src/observe/meta/monitoring/schedule.go#83). So something needs to change in the codebase first (either making the scheduling optional when trying to update a monitor, or having the GraphQL return the default scheduling when fetching a monitor instead of null)

obs-gh-abhinavpappu:
I was wondering why updates from the UI were still working even though they also didn't contain a `scheduling` block by default, and turns out the API is already supposed to treat `scheduling` as optional for updates. However, the terraform provider was sending `{scheduling: {interval: null, transform: null}}` instead of `{scheduling: null}` like the UI, and the API checks whether `scheduling` is null to determine whether to use the default transform option.

Updated to send `{scheduling: null}` like the UI

vikramraman (approved):

PR #151, opened 2024-08-29
Remove schedule block from monitorv2

obs-gh-maxhahn:
This feature had been moved upstream and no longer exists in this resource.


vikramraman:
Need to also remove it from data_source_monitor_v2.go

vikramraman (approved):

PR #159, opened 2024-10-07
Add binding export functionality to dashboards

obs-gh-maxhahn:
This is an internal use options that allows id references to be collected in a unified location in the resulting datasource read schema for further postprocessing.

### DIFF
--- a/client/binding/binding.go
+++ b/client/binding/binding.go
@@ -5,10 +5,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 
 	observe "github.com/observeinc/terraform-provider-observe/client"
 	"github.com/observeinc/terraform-provider-observe/client/meta/types"
+	"github.com/observeinc/terraform-provider-observe/client/oid"
 )
 
 var (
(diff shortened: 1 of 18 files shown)


observe/resource_dashboard.go
```diff
@@ -119,7 +120,8 @@ func newDashboardConfig(data *schema.ResourceData) (input *gql.DashboardInput, d
 	}
 
 	if v, ok := data.GetOk("layout"); ok {
-		input.Layout = types.JsonObject(v.(string)).Ptr()
+		layoutJson := types.JsonObject(v.(string)).Ptr()
+		input.Layout = layoutJson.Ptr()
```
obs-gh-abhinavpappu, line 124: revert to `input.Layout = types.JsonObject(v.(string)).Ptr()`

client/binding/binding.go
```diff
@@ -0,0 +1,279 @@
...
+			worksheets, err := client.ListWorksheetIdLabelOnly(ctx, optArgs.workspaceId)
+			if err != nil {
+				return cache, err
+			}
+			for _, wk := range worksheets {
+				cache.addEntry(KindWorksheet, wk.Label, wk.Id, &disambiguator, existingResourceNames)
+			}
+		case KindWorkspace:
+			// special case: one workspace per customer, will always be referred to as
+			// "default"
```
obs-gh-abhinavpappu, line 52: update to use actual workspace name

client/binding/binding.go
```diff
@@ -0,0 +1,279 @@
...
+				b := Ref{kind: kind, key: maybeLabel.Value}
+				terraformLocal := sanitizeIdentifier(fmt.Sprintf("binding__%s_%s__%s_%s", g.resourceType, g.resourceName, kind, maybeLabel.TfName))
+				g.bindings[b] = Target{
+					Value:             value,
+					TfName:            maybeLabel.TfName,
+					TfLocalBindingVar: terraformLocal,
+				}
+				jsonMapNode[key] = fmt.Sprintf("${local.%s}", terraformLocal)
+				break
+			}
```
obs-gh-abhinavpappu, line 162: could this be replaced with TryBind?

obs-gh-abhinavpappu (commented):
Synced with Max offline, overall looks good! Should be good to go after adding some tests

obs-gh-maxhahn:
All feedback points should be addressed and this patch should be ready for review

observe/resource_dashboard.go
```diff
@@ -200,11 +208,36 @@ func dashboardToResourceData(d *gql.Dashboard, data *schema.ResourceData) (diags
...
 			diags = append(diags, diag.FromErr(diagErr)...)
+		} else if parameterValuesRaw, err := gen.GenerateJson(parameterValuesRaw); err != nil {
+			diagErr := fmt.Errorf("failed to generate bindings for 'parameterValuesRaw' response field: %w", err)
+			diags = append(diags, diag.FromErr(diagErr)...)
 		} else if err := data.Set("parameter_values", string(parameterValuesRaw)); err != nil {
 			diags = append(diags, diag.FromErr(err)...)
 		}
 	}
 
+	if d.Layout != nil || gen.HasBindings() {
```
obs-gh-abhinavpappu, line 219: nit: The `gen.HasBindings()` here is really just checking that `g.enabled` is true right? Since the other check of `len(g.bindings) > 0` will always be true if `g.enabled` is true, due to the `TryBind` above for workspace. Maybe just checking if it's enabled would be better since checking `gen.HasBindings()` seems to imply that the layout bindings may have already been generated.

obs-gh-abhinavpappu (approved):
overall lgtm!

observe/resource_dashboard.go
obs-gh-maxhahn: hrm, good point. Given that we always will try bind a workspace, and it shouldn't fail, I agree that switching to enabled will be clearer.

PR #160, opened 2024-10-08
feat: add new observe_grant resource

obs-gh-abhinavpappu:
Create a new `observe_grant` resource for working with RBAC v2 statements. For now, this resource translates to and from the v1 statement format until API support is added for grants.


observe/resource_grant.go
```diff
@@ -0,0 +1,364 @@
...
+		if err != nil {
+			return nil, diag.Errorf("error parsing qualifier.oid: %s", err.Error())
+		}
+		resourceId = &resourceOid.Id
+	}
+	input.Object, err = role.ToRbacObject(resourceId)
+	if err != nil {
+		return nil, diag.Errorf(err.Error())
+	}
+	input.Object.Owner = boolPtr(false)
```
vikramraman, line 109: Could these bits just be handled within the `ToRbacObject` func?

vikramraman (approved):

PR #162, opened 2024-10-15
feat: Add correlation tags to dataset datasource schema

obs-gh-maxhahn:
This should be exposed for downstream consumption so users can obtain the set of correlation tags currently attached to their datasets.

### DIFF
--- a/docs/data-sources/dataset.md
+++ b/docs/data-sources/dataset.md
@@ -28,7 +28,6 @@ data "observe_datastet" "example" {
 
 ### Optional
 
-- `correlation_tag` (Block List) Correlation tags associated with this dataset. (see [below for nested schema](#nestedblock--correlation_tag))
 - `id` (String) Resource ID for this object.
 One of `name` or `id` must be set.
 - `name` (String) Dataset name. Must be unique within workspace.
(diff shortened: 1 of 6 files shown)


docs/data-sources/dataset.md
```diff
@@ -54,6 +55,29 @@ paths between two datasets.
 input is provided, a stage will implicitly follow on from the result of
 its predecessor. (see [below for nested schema](#nestedblock--stage))
 
+<a id="nestedblock--correlation_tag"></a>
+### Nested Schema for `correlation_tag`
+
+Required:
```
vikramraman, line 61: All of this should be marked as read-only? This is not required for the data lookups.

observe/data_source_dataset.go
```diff
@@ -156,5 +185,17 @@ func dataSourceDatasetRead(ctx context.Context, data *schema.ResourceData, meta
...
+	if d.CorrelationTagMappings != nil {
+		var cts []interface{}
+		for _, ct := range d.CorrelationTagMappings {
+			cts = append(cts, map[string]interface{}{
+				correlationTagNameKey:   ct.Tag,
+				correlationTagColumnKey: ct.Path.Column,
+				correlationTagPathKey:   ct.Path.Path,
+			})
+		}
+		data.Set("correlation_tag", cts)
```
nitishch, line 198: How are you planning to use this correlation_tag as part of dataset export?

docs/data-sources/dataset.md
obs-gh-maxhahn: Ah thanks, forgot to set the computed attribute.

obs-gh-maxhahn: Thanks, forgot to tag the subfield with the computed attrib.

observe/data_source_dataset.go
obs-gh-maxhahn: We can grab this from the exporter side and convert it to actual correlation_tag resources.

vikramraman (approved):

PR #166, opened 2024-10-23
feat: Allow edit-forward of Terraform-managed datasets

obs-gh-jamesweng:
Expose `saveDataset`'s `rematerialization_mode` to `resource_dataset` in Terraform. Must be merged after 10/30 prod release because GraphQL API not released yet. Tested and verified with `testbox`.

### DIFF
--- a/observe/descriptions/dataset.yaml
+++ b/observe/descriptions/dataset.yaml
@@ -18,3 +18,7 @@ schema:
   correlation_tag:
     description: |
       Correlation tags associated with this dataset.
+  rematerialization_mode: |
+    Specifies rematerialization mode when updating a dataset. Options include
+    "rematerialize" and "skip_rematerialization". If no option is used, "rematerialize"
+    is used by default.
--- a/observe/descriptions/dataset.yaml
+++ b/observe/descriptions/dataset.yaml
@@ -20,5 +20,6 @@ schema:
       Correlation tags associated with this dataset.
   rematerialization_mode: |
     Specifies rematerialization mode when updating a dataset. Options include
-    "rematerialize" and "skip_rematerialization". If no option is used, "rematerialize"
-    is used by default.
+    "rematerialize" and "skip_rematerialization" - if no option is used, "rematerialize"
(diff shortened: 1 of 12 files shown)


observe/descriptions/dataset.yaml
```diff
@@ -18,3 +18,7 @@ schema:
   correlation_tag:
     description: |
       Correlation tags associated with this dataset.
+  rematerialization_mode: |
```
obs-gh-abhinavpappu, line 21: maybe worth mentioning here that skip_rematerialization doesn't guarantee that the dataset won't be rematerialized if i'm understanding this correctly

observe/resource_dataset.go
```diff
@@ -143,6 +148,13 @@ func resourceDataset() *schema.Resource {
 					},
 				},
 			},
+			"rematerialization_mode": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          "rematerialize",
+				ValidateDiagFunc: validateRematerializationMode,
```
obs-gh-abhinavpappu, line 155: can use `validateEnums` from `observe/helpers.go` instead and then use the gql generated enums directly instead of the constants above, see examples from `resource_monitor_v2.go`

obs-gh-abhinavpappu (commented):
Overall looks great, though it's a bit weird because it seems the rematerialization mode is not really a "state" that can be saved for the dataset, but rather something that only applies to the update action itself. But I don't know if terraform has a better way to deal with things like this, so this might be fine, would like @vikramraman to take a quick look as well before approving

observe/descriptions/dataset.yaml
obs-gh-jamesweng: Good catch, done.

observe/resource_dataset.go
obs-gh-jamesweng: Done.

obs-gh-abhinavpappu (approved):
talked to Vikram, this looks good!

PR #169, opened 2024-11-08
Update dataset version field to use lastSaved instead of updatedDate

obs-gh-maxhahn:
This field is the proper one to use to determine when the last user modifications were made to a dataset which also excludes system modifications and is required to properly support an upcoming release


client/internal/meta/schema/base.graphql
vikramraman, line 1: Could we rope in only the gql changes you need for this PR?

We can update the rest in a separate PR

vikramraman (changes requested):

vikramraman (approved):

PR #170, opened 2024-11-08
Fix failing dataset tests

obs-gh-maxhahn:
It is now illegal to convert between materialized datasets and view datasets, so we must update the test accordingly.


obs-gh-abhinavpappu (approved):
lgtm, thanks for fixing it!

PR #172, opened 2024-11-18
feat: support targeting Everyone group before RBAC v2 is turned on

obs-gh-abhinavpappu:


observe/data_source_rbac_group.go
```diff
@@ -65,6 +66,18 @@ func dataSourceRbacGroupRead(ctx context.Context, data *schema.ResourceData, met
 		r, err = client.LookupRbacGroup(ctx, name)
 	}
 
+	// The Everyone group only exists if RBAC v2 is turned on for the customer.
+	// However, we want to support creating RBAC v2 statements before it's turned on,
+	// so we return the special Everyone group id 1 if name is "Everyone" and there
+	// is no existing v1 group called "Everyone".
+	if err != nil && name == "Everyone" {
```
vikramraman, line 73: Can we do this only if the error is 404 equivalent and not for all errors?

obs-gh-abhinavpappu: updated

vikramraman (approved):

PR #173, opened 2024-11-19
feat: add authoritative resource to manage grants for an object

obs-gh-abhinavpappu:
Ensures there are no grants for the specified object other than the ones specified in terraform.

I'm using a `schema.TypeSet`, which is really convenient because it associates existing statements with the ones defined in terraform automatically based on the `subject` and `role` regardless of the statement id. So for example, say terraform defines statements A and B which get created with ids 8001 and 8002. Then if those statements _manually_ get deleted, and then new statements C and D get created with the same `subject` and `role`, now with ids 8003 and 8004, `terraform apply` will make no changes. Otherwise, we create and delete the minimal number of statements to make it match (no statements are updated since we want them to be immutable with v2).


obs-gh-abhinavpappu:
Also thoughts on renaming this to something like `observe_object_grants` or `observe_resource_grants`? I think `observe_grants` might be too similar to `observe_grant`, and we may want to add something like `observe_group_grants`/`observe_role_grants` in the future that's the authoritative resource to manage `role` grants for a group.

observe/resource_grants.go
```diff
@@ -0,0 +1,209 @@
...
+				Type:             schema.TypeString,
+				Required:         true,
+				ValidateDiagFunc: validateOID(validRbacV2Types...),
+				DiffSuppressFunc: diffSuppressOIDVersion,
+				Description:      descriptions.Get("grants", "schema", "oid"),
+				ForceNew:         true,
+			},
+			"grant": {
+				Type:     schema.TypeSet,
+				Optional: true,
```
vikramraman, line 35: Do we enforce somewhere that at least one of this must be specified?

vikramraman:
> Also thoughts on renaming this to something like `observe_object_grants` or `observe_resource_grants`? I think `observe_grants` might be too similar to `observe_grant`, and we may want to add something like `observe_group_grants`/`observe_role_grants` in the future that's the authoritative resource to manage `role` grants for a group.

`observe_resource_grants` sounds good

observe/resource_grants.go
obs-gh-abhinavpappu: No. I was considering it, but I think it makes sense to say "I want no grants for this object" (i.e. only admins can access it). I don't know if doing the following
```hcl
resource "observe_resource_grants" "example" {
    oid = data.observe_dataset.super-secret.oid
}
```
actually makes it clear that only admins can access it, but seems like a valid use case.

obs-gh-abhinavpappu: would it be better to introduce another field like `admin_only` and enforce that either `admin_only = true` or at least one grant is specified?

vikramraman: Ah interesting. No, I think omitting grants to suggest only admins can do it is fine.
Would be good to call that out in the examples, description so people are aware of the implications there.

obs-gh-abhinavpappu:
Still thinking of ways to rename `observe_resource_grants` it so it's clearer, looked at some other products' terraform resources, but I don't think they're particularly clear either.

Also not sure if this would help, but I'm open to wrap the `oid` in a `target` block or something like that:
```
resource "observe_resource_grants" "example" {
    target {
        oid = data.observe_dataset.example.oid
    }
    ...
}
```
I don't want to use `qualifier` even though having the name match `observe_grant` would be nice, since it doesn't make as much sense here. Using `qualifier` sounds like I'm defining a bunch of grants and then just qualifying them with an object id, which is true, but doesn't convey that the object id is more important in this case, with all existing statements for that object being replaced.

docs/resources/resource_grants.md
```diff
@@ -0,0 +1,91 @@
...
+description: |-
+  NOTE: This feature is still under development. It is not meant for customer use yet.
+  Authoritative. Manages the complete set of grants for a given resource, replacing
+  any existing ones. If no grants are specified, only admins will have access.
+---
+# observe_resource_grants
+
+NOTE: This feature is still under development. It is not meant for customer use yet.
+
+Authoritative. Manages the complete set of grants for a given resource, replacing
```
vikramraman, line 14: Can we also add a comment that this should not be used with `observe_grant` for the same resource.

vikramraman:
> Still thinking of ways to rename `observe_resource_grants` it so it's clearer, looked at some other products' terraform resources, but I don't think they're particularly clear either.
> 
> Also not sure if this would help, but I'm open to wrap the `oid` in a `target` block or something like that:
> 
> ```
> resource "observe_resource_grants" "example" {
>     target {
>         oid = data.observe_dataset.example.oid
>     }
>     ...
> }
> ```
> 
> I don't want to use `qualifier` even though having the name match `observe_grant` would be nice, since it doesn't make as much sense here. Using `qualifier` sounds like I'm defining a bunch of grants and then just qualifying them with an object id, which is true, but doesn't convey that the object id is more important in this case, with all existing statements for that object being replaced.

Is there a reason to introduce a target block here? I'm fine just leaving oid as a top level for now.

docs/resources/resource_grants.md
obs-gh-abhinavpappu: done

vikramraman (approved):

PR #174, opened 2024-11-20
feature: update monitorv2 to allow inline actions

obs-gh-catherman:
We moved the internal API to support one-shot create/update call SaveMonitorV2WithActions. This migrates the existing monitorv2 resource to use that function, which also allows for specifying shared actions (current feature support) and now also inline actions. We require that shared actions be separately managed resources and set using action id (again, current feautre support). Action ID is now optional, and instead a full action definition can be given inline which we translate to an Inline:true (private action).


observe/resource_monitor_v2.go
```diff
@@ -524,6 +570,12 @@ func resourceMonitorV2Read(ctx context.Context, data *schema.ResourceData, meta
 		}
 	}
 
+	if monitor.Definition.MaxAlertsPerHour != nil {
+		if err := data.Set("max_alerts_per_hour", monitor.Definition.MaxAlertsPerHour.String()); err != nil {
```
obs-gh-abhinavpappu, line 574: does this not need to be converted to an int64? maybe the sdk does it automatically, but could you add this field to the test case to confirm

observe/resource_monitor_v2.go
```diff
@@ -577,20 +629,47 @@ func monitorV2FlattenRule(gqlRule gql.MonitorV2Rule) interface{} {
...
+		// This is a shared action. It is assumed the monitor was created with the action id set
+		// and so that's what we return in the resource.
+		rules["oid"] = oid.MonitorV2ActionOid(gqlActionRule.ActionID).String()
+	} else {
+		actMap := map[string]any{}
+		rules["action"] = []any{actMap}
+
+		// This is an inline/private action. We don't (yet?) get everything we want to populate in the
+		// response so we have to do a read to get this.
+		action, err := client.GetMonitorV2Action(ctx, oid.MonitorV2ActionOid(gqlActionRule.ActionID).String())
```
obs-gh-abhinavpappu, line 653: Would be good to create a low-priority ticket to include the missing information in the `getMonitorV2` response (i think it's just `description` that's missing?), so we can avoid extra api calls. I'm sure the frontend would also find that useful.

observe/resource_monitor_v2.go
```diff
@@ -770,6 +849,22 @@ func monitorV2FlattenTransformSchedule(gqlTransformSchedule gql.MonitorV2Transfo
 	return []interface{}{transformSchedule}
 }
 
+func newMonitorV2WithActions(data *schema.ResourceData) (actions []gql.MonitorV2ActionAndRelationInput, diags diag.Diagnostics) {
```
obs-gh-abhinavpappu, line 852: nit: `newMonitorV2WithActions` makes it sound like this function creates both the monitor input and the action input, when it actually only does the actions, maybe `newMonitorV2Actions`/`newMonitorV2ActionRelations` instead?

observe/resource_monitor_v2_action.go
```diff
@@ -36,6 +36,11 @@ func resourceMonitorV2Action() *schema.Resource {
 				ValidateDiagFunc: validateEnums(gql.AllMonitorV2ActionTypes),
 				Required:         true,
 			},
+			// TODO: remove this. we are only doing this for shared actions
+			"inline": { // Boolean
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
```
obs-gh-abhinavpappu, line 43: remove? along with corresponding code in `newMonitorV2ActionInput`

observe/resource_monitor_v2_action_test.go
```diff
@@ -348,3 +348,93 @@ func TestAccObserveMonitorV2MultipleActionsEmail(t *testing.T) {
 		},
 	})
 }
+
+func TestAccObserveMonitorV2MultipleActionsEmailViaOneShot(t *testing.T) {
```
obs-gh-abhinavpappu, line 352: i think it makes more sense to put this test in `resource_monitor_v2_test.go` since these actions are set on an `observe_monitor_v2` resource

observe/resource_monitor_v2_action_test.go
```diff
@@ -348,3 +348,93 @@ func TestAccObserveMonitorV2MultipleActionsEmail(t *testing.T) {
...
+							send_end_notifications = false
+							send_reminders_interval = "20m"
+						}
+					}
+
+					data "observe_user" "system" {
+						email = "%[2]s"
+					}
+				`, randomPrefix, systemUser()),
+				Check: resource.ComposeTestCheckFunc(),
```
obs-gh-abhinavpappu, line 436: add checks :) doesn't need to be exhaustive since the other tests already check for a lot of things, but some basic ones for the actions would be nice

observe/resource_monitor_v2_action_test.go
```diff
@@ -348,3 +348,93 @@ func TestAccObserveMonitorV2MultipleActionsEmail(t *testing.T) {
...
+			},
```
obs-gh-abhinavpappu, line 437: Would be good to test that `update` is working as well. To do so, you can create a new test step, copy paste the `config` from the first step, and change some of the fields for the inline actions. And add checks to ensure those fields were actually updated as desired.

The test works by essentially running the first step (which would be a create since there's a new resource in the config), then replacing the terraform config from the first step with the config from the second step while preserving the `tfstate`, so any changes made in the second step to existing resources are updates.

obs-gh-abhinavpappu (commented):
Left some comments, but overall looks good!

observe/resource_monitor_v2.go
obs-gh-catherman: OB-39564

observe/resource_monitor_v2.go
obs-gh-catherman: I'm not sure what I would need to do for that. This is just setting an unquoted string, right?

obs-gh-catherman: nm. i see it in a test fail.

obs-gh-abhinavpappu (approved):
lgtm!

PR #175, opened 2024-12-02
fix: always diff suppress rematerialization_mode

obs-gh-abhinavpappu:


obs-gh-abhinavpappu:
Never mind, this doesn't do what I intended. It not only prevents a diff when this field is updated (which is what I want), but also it appears that when other fields (such as the OPAL) are updated, any changes to `rematerialization_mode` just aren't considered at all (i.e. you can essentially never change the `rematerialization_mode` after creation)

PR #176, opened 2024-12-03
fix: prevent unnecessary diff for rematerialization_mode

obs-gh-abhinavpappu:
Currently, for any existing dataset resources, the value for `rematerialization_mode` in the tfstate file is `""` since unlike for other fields that are set in `resourceDatasetRead`, `rematerialization_mode` isn't an actual property of datasets and there's no value for it returned by the API. This results in an unnecessary diff of `"" -> "rematerialize"` even though `rematerialize` is the default value.

To avoid this, I updated `resourceDatasetRead` to set `rematerialization_mode` to `rematerialize` if there's not already a value for it.

### DIFF
--- a/observe/resource_dataset.go
+++ b/observe/resource_dataset.go
@@ -21,9 +21,6 @@ const (
 	schemaDatasetDescriptionDescription = "Dataset description."
 	schemaDatasetIconDescription        = "Icon image."
 	schemaDatasetOIDDescription         = "The Observe ID for dataset."
-
-	rematerializationModeRematerialize         = "rematerialize"
-	rematerializationModeSkipRematerialization = "skip_rematerialization"
 )
 
 func resourceDataset() *schema.Resource {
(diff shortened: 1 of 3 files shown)


observe/resource_dataset.go
```diff
@@ -406,6 +404,14 @@ func resourceDatasetRead(ctx context.Context, data *schema.ResourceData, meta in
 		})
 	}
 
+	// Since rematerialization_mode is not an actual property of the dataset, we don't have a value
+	// for it in the `result` from the API. So we set it to "rematerialize" (the default value)
```
vikramraman, line 408: This change is good to get in! Just curious whether there's a way to get what the default is via the API? 

Otherwise if the default changes it would break TF until we update again.

vikramraman (approved):

observe/resource_dataset.go
obs-gh-abhinavpappu: The other option to prevent breaking anything if the API changes its default would be to keep the existing options (`rematerialize` or `skip_rematerialization`) and always explicitly set the value in the API call. Then if the API changes its default, terraform would continue to use `rematerialize` by default (until the provider is updated) but it wouldn't break anything.

obs-gh-abhinavpappu: > have the default value for rematerialization_mode be "" instead of "rematerialize"

Updated it to this. Unfortunately this will mean that anyone who ran terraform apply since `rematerialization_mode` was added will see another unnecessary diff, but I still think this is probably the best approach.

PR #178, opened 2024-12-09
feat: Add conditions to monv2 action rules

obs-gh-catherman:


observe/resource_monitor_v2.go
```diff
@@ -299,6 +299,41 @@ func resourceMonitorV2() *schema.Resource {
...
+							Optional:    true,
+							MaxItems:    1,
+							Description: descriptions.Get("monitorv2", "schema", "actions", "conditions", "description"),
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									// note: subExpressions will be interesting to support. The UI currently does
+									// not support it, so we don't here either.
+									"compare_terms": { // [MonitorV2ComparisonTerm!]
+										Type:     schema.TypeList,
+										Optional: true,
```
obs-gh-abhinavpappu, line 313: oh wait this should probably be required with minitems 1

obs-gh-catherman: oh, good point. until i add support for sub-expressions, it is a minitems=1 instead of a exactlyoneof kind of thing, huh?

PR #180, opened 2024-12-16
fix: add inline actions and conditions to monitor v2 data source to support exporting

obs-gh-abhinavpappu:


observe/resource_monitor_v2.go
```diff
@@ -35,18 +35,18 @@ func resourceMonitorV2() *schema.Resource {
 				ValidateDiagFunc: validateOID(oid.TypeWorkspace),
 				Description:      descriptions.Get("monitorv2", "schema", "workspace_id"),
 			},
+			"name": { // String!
```
obs-gh-maxhahn, line 38: Was this diff meant to fix something?

obs-gh-abhinavpappu: oh oops, i switched those two around to make it match the order of the fields in the data source, so that the files are easy to compare side-by-side and make sure i don't miss anything, meant to switch them back but i think i'll just leave them since it doesn't really affect anything

obs-gh-maxhahn (approved):

vikramraman (approved):

PR #184, opened 2025-01-14
feat: Add notification_fragments to monv2 definition

obs-gh-catherman:


obs-gh-abhinavpappu (commented):
This looks good to me, not sure why the test is failing. Has the API change been merged in yet?

PR #185, opened 2025-01-18
Fix two bugs in the resource_monitor_v2

obs-gh-andrewlee:
1. DiffSuppressFunc set to diffSuppressTimeDuration for lookback_time. This is to ensure a null lookback_time doesn't trigger the resource to update (check: OB-40067).

2. value_duration Type set to schema.TypeString as the validateTimeDuration would panic when casting the type integer to a string.

### DIFF
--- a/docs/resources/monitor_v2.md
+++ b/docs/resources/monitor_v2.md
@@ -81,7 +81,7 @@ Required:
 Optional:
 
 - `value_bool` (List of Boolean) list of size <=1 consisting of a boolean value.
-- `value_duration` (List of Number) list of size <=1 consisting of a duration value.
+- `value_duration` (List of String) list of size <=1 consisting of a duration value.
 - `value_float64` (List of Number) list of size <=1 consisting of a float value.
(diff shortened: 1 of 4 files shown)


observe/resource_rbac_statement_test.go
vikramraman, line 1: Can you rebase against the main branch. Just merged Abhi's changes skipping these tests as well.
So this is no longer needed here.

obs-gh-andrewlee: sgtm 💯

obs-gh-catherman (commented):
Does this work backward compatibly with existing definitions? Seems like no.

obs-gh-andrewlee:
> Does this work backward compatibly with existing definitions? Seems like no.

This is backward compatible because there are no monitors in prod that uses the value_duration field. If so, their terraform would've never have succeeded in being applied.

obs-gh-catherman (approved):
big fan. let's get abhi to also look.

obs-gh-abhinavpappu (approved):
lgtm!

vikramraman (approved):

PR #196, opened 2025-02-07
feat: add initial support for reference tables

obs-gh-abhinavpappu:
- move current REST client (only used for login) from `client/internal/customer` to `client/rest`
- add helper functions for POST, GET, PUT, PATCH, and DELETE
    - return error type including status code, allowing `resourceReferenceTableRead` to determine whether the reference table exists (code 404)
- add types and functions to call reference table API - will be mostly generated in a future PR
- add new resource and data source `observe_reference_table`
    - To avoid storing the entire (potentially very large) file in the state, as well as needing to fetch the entire file on every `plan/refresh/apply` to check for changes, we use a `checksum` field to detect changes instead. This `checksum` should always be set to `filemd5("<path_to_source_file>")`. The API returns the current remote checksum, which is compared against the local one.
    - There are two ways to update a reference table, PUT and PATCH. PUT will replace everything with what's provided, and PATCH will update only the metadata fields that are provided in the request, leaving the others as is. PATCH can't be used to update the file itself.
- add basic tests using some test csv files added to `observe/testdata`

**In future PRs, will:**
- Generate types (and functions?) from the OpenAPI spec using [openapi-generator](https://github.com/OpenAPITools/openapi-generator)
- A few other minor changes after API is updated, indicated by TODO comments


client/rest/referencetables.go
```diff
@@ -0,0 +1,184 @@
...
+	"fmt"
+	"io"
+	"mime/multipart"
+	"os"
+	"path/filepath"
+
+	"github.com/observeinc/terraform-provider-observe/client/oid"
+)
+
+// TODO: generate types from OpenAPI spec
```
vikramraman, line 16: I suspect most of this file can be auto-generated from the spec once we settle on the mechanics

docs/resources/reference_table.md
```diff
@@ -0,0 +1,43 @@
...
+  A reference table represents a source of non-temporal data being ingested into Observe.
+---
+# observe_reference_table
+
+A reference table represents a source of non-temporal data being ingested into Observe.
+## Example Usage
+```terraform
+resource "observe_reference_table" "example" {
+  label       = "Example"
+  source      = "path/to/reference_table.csv"
```
vikramraman, line 15: Could we rename this to `source_file` for clarity.

docs/data-sources/reference_table.md
```diff
@@ -0,0 +1,31 @@
...
+A reference table represents a source of non-temporal data being ingested into Observe.
+
+
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Optional
+
+- `id` (String) Resource ID for this object.
```
vikramraman, line 20: I assume we need one of id or label? If so, let's make that clear in the description.

observe/resource_reference_table.go
```diff
@@ -0,0 +1,222 @@
...
+	input = &rest.ReferenceTableInput{
+		Metadata:       *metadataInput,
+		SourceFilePath: data.Get("source").(string),
+	}
+
+	return input, diags
+}
+
+func newReferenceTableMetadataConfig(data *schema.ResourceData, patch bool) (input *rest.ReferenceTableMetadataInput, diags diag.Diagnostics) {
+	// If we're using PATCH, then we only want to set fields that have been modified.
```
vikramraman, line 96: Curious if this matters for terraform. We always have the entire state, so we can likely always use PUT to keep it simple?

observe/resource_reference_table.go
```diff
@@ -0,0 +1,222 @@
...
+
+	if err := data.Set("dataset", oid.DatasetOid(d.DatasetId).String()); err != nil {
+		diags = append(diags, diag.FromErr(err)...)
+	}
+
+	if err := data.Set("checksum", d.Checksum); err != nil {
+		diags = append(diags, diag.FromErr(err)...)
+	}
+
+	// TODO: add "primary_key" and "label_field" once supported
```
vikramraman, line 139: Not sure if you meant to add it in this PR as we do support it? Or do we need support on the api front for this?

observe/resource_reference_table.go
```diff
@@ -0,0 +1,222 @@
...
+		return diags
+	}
+
+	return append(diags, resourceReferenceTableRead(ctx, data, meta)...)
+}
+
+func resourceReferenceTableDelete(ctx context.Context, data *schema.ResourceData, meta interface{}) (diags diag.Diagnostics) {
+	client := meta.(*observe.Client)
+	if err := client.DeleteReferenceTable(ctx, data.Id()); err != nil {
+		return diag.Errorf("failed to delete reference table: %s", err)
```
vikramraman, line 219: If the resource was already deleted server side (404), we would want to just remove it from state rather than throw an error.

vikramraman (changes requested):

observe/resource_reference_table.go
obs-gh-abhinavpappu: We could, but this would mean re-uploading the (potentially very large) file even if it hasn't changed

observe/resource_reference_table.go
obs-gh-abhinavpappu: Need API support. Currently it doesn't return `primary_key` or `label_field` in the response, but they will be added soon. So currently, we're unable to detect changes to those fields made outside of terraform, but it should work fine as long as all changes are made through terraform. Don't think this is a blocker since there isn't UI support for editing reference tables, so the only way to change those fields outside of terraform is through the API (which is unlikely to happen for a terraform-managed resource).

observe/resource_reference_table.go
vikramraman: Got it. Using patch works.

observe/resource_reference_table.go
obs-gh-abhinavpappu: In most cases, this wouldn't be an issue since Terraform always runs `read` before `delete`, and the `read` would remove it from the state, resulting in terraform detecting no changes and never running `delete`.

Did add the check though since the resource can be deleted between when the `read` and `delete` are run. And there can be a long gap between them by running `terraform apply` with a previously saved plan (or just by waiting before typing in `"yes"` to apply changes)

client/rest/referencetables.go
obs-gh-abhinavpappu: Yep, got something working for generation and this file can be removed entirely. But requires some OpenAPI spec changes that are currently still in progress, so will add in a separate PR once those are in

vikramraman (approved):

PR #197, opened 2025-02-11
feat: add schema support for reference tables

obs-gh-abhinavpappu:
Needed to use `data.GetRawConfig()` to work around a terraform limitation, see comment for details.

### DIFF
--- a/observe/resource_reference_table.go
+++ b/observe/resource_reference_table.go
@@ -141,13 +141,25 @@ func newReferenceTableMetadataConfig(data *schema.ResourceData, patch bool) (inp
 }
 
 func newReferenceTableSchemaConfig(data *schema.ResourceData) (input []rest.ReferenceTableSchemaInput, diags diag.Diagnostics) {
-	schema := data.Get("schema").([]interface{})
-	for _, v := range schema {
-		s := v.(map[string]interface{})
-		input = append(input, rest.ReferenceTableSchemaInput{
-			Name: s["name"].(string),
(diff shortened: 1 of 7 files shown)


observe/resource_reference_table.go
```diff
@@ -116,24 +140,60 @@ func newReferenceTableMetadataConfig(data *schema.ResourceData, patch bool) (inp
 	return input, diags
 }
 
-func referenceTableToResourceData(d *rest.ReferenceTable, data *schema.ResourceData) (diags diag.Diagnostics) {
-	if err := data.Set("oid", d.Oid().String()); err != nil {
+func newReferenceTableSchemaConfig(data *schema.ResourceData) (input []rest.ReferenceTableSchemaInput, diags diag.Diagnostics) {
+	// We're using GetRawConfig() here instead of Get() or GetOk() because we only want to include
```
vikramraman, line 144: I wonder if we could leverage `PlanModifiers` instead: https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#attribute-plan-modification

This would be configured on the schema definition for the field, and allows us to modify the plan as relevant.

obs-gh-abhinavpappu: Looks like `PlanModifiers` are only supported for the Framework and not the SDK. But yes, it does give us what we need to implement this behavior (and in fact from what I can tell based on the docs, I think this behavior is default in the Framework)

vikramraman (approved):

PR #201, opened 2025-02-20
feat: add disabled flag support for monitors

obs-gh-catherman:


observe/resource_monitor_v2.go
```diff
@@ -119,7 +124,7 @@ func resourceMonitorV2() *schema.Resource {
 							Description: descriptions.Get("monitorv2", "schema", "rules", "count"),
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
-									"compare_values": { // [MonitorV2ComparisonInput!]!
+									"g": { // [MonitorV2ComparisonInput!]!
```
obs-gh-abhinavpappu, line 127: this doesn't look intentional

obs-gh-abhinavpappu (changes requested):
please add it to the data source as well (necessary to support export), and would be nice to include it in the test cases somewhere

obs-gh-abhinavpappu:
> please add it to the data source as well

bump in case you missed this comment :)

observe/resource_monitor_v2.go
```diff
@@ -952,6 +962,9 @@ func newMonitorV2Input(data *schema.ResourceData) (input *gql.MonitorV2Input, di
 	if v, ok := data.GetOk("description"); ok {
 		input.Description = stringPtr(v.(string))
 	}
+	if v, ok := data.GetOk("disabled"); ok {
+		input.Disabled = boolPtr(v.(bool))
+	}
```
obs-gh-abhinavpappu, line 967: Due to a terraform sdk limitation, if `disabled` is `false`, then the `ok` here will always be `false` (and therefore `input.Disabled` will be `nil`/omitted instead of being `false`). So assuming that the API treats `disabled` being omitted in the input as "leave unchanged", this would mean it's impossible re-enable a monitor through terraform after disabling it. Would probably suggest doing the following instead:
```suggestion
	input.Disabled = boolPtr(data.Get("disabled").(bool))
```

obs-gh-abhinavpappu: Actually it looks like the API treats `disabled` being omitted as `false`, so approving. But may want to make this change anyway

obs-gh-abhinavpappu (approved):

PR #204, opened 2025-02-25
feat: Fail terraform apply if edit-forward cannot be applied

obs-gh-jamesweng:
Currently if a dataset cannot be edited-forward, it will be rematerialized (best-effort behavior). This is an unpleasant surprise for users because there is no mechanism to perform dry-run in Terraform (unlike the WebUI). This change fails those changes, and the user should either make their change compatible with edit-forward or remove the rematerialization_mode flag.


obs-gh-jamesweng:
It seems some RBAC tests are failing - I don't think my change would've caused that but I'm not 100% sure.

danielodievich:
Would this fail in the app? 

I have a bunch of datasets in the o4s app (https://github.com/observeinc/terraform-observe-snowflake/) and I would like them to be edit-forward so I am doing this:

```terraform
resource "observe_dataset" "alert_history" {
  workspace              = var.workspace.oid
  name                   = format(var.name_format, "ALERT_HISTORY")
  icon_url               = "time-and-date/calendar-operations/calendar-minus"
  freshness              = lookup(var.freshness_overrides, "alert_history", var.freshness_default_duration)
  rematerialization_mode = "skip_rematerialization"
  description            = <<-EOF
    Alert usage history (SNOWFLAKE.ACCOUNT_USAGE.ALERT_HISTORY view)
  EOF
```

Now I make an update to the app, and am making changes incompatible with edit_forward. 
Is my app upgrade hosed?

observe/resource_dataset.go
```diff
@@ -421,6 +422,35 @@ func resourceDatasetUpdate(ctx context.Context, data *schema.ResourceData, meta
...
+		if result, err := client.SaveDatasetDryRun(ctx, wsid.Id, input, queryInput); err != nil {
+			diags = append(diags, diag.Diagnostic{
+				Severity: diag.Error,
+				Summary:  fmt.Sprintf("failed to update dataset [id=%s]", data.Id()),
+				Detail:   err.Error(),
+			})
+			return diags
+		} else if len(result) > 0 {
+			var sb strings.Builder
+			sb.WriteString("The following dataset(s) will be rematerialized: ")
```
obs-gh-abhinavpappu, line 437: nit: maybe include something indicating that if they're ok with rematerializing the datasets, to get rid of the `rematerialization_mode = skip_rematerialization` and try again

client/meta/dataset.go
```diff
@@ -36,6 +36,22 @@ func (client *Client) SaveDataset(ctx context.Context, workspaceId string, input
 	return datasetOrError(resp.Dataset, err)
 }
 
+// SaveDatasetDryRun saves a dataset with pre-flight checks - this is useful for
```
obs-gh-abhinavpappu, line 39: finish comment

observe/resource_dataset_test.go
```diff
@@ -129,7 +129,6 @@ func TestAccObserveDatasetUpdate(t *testing.T) {
 					}
 
 					data_table_view_state = jsonencode({viewType = "Auto"})
-					rematerialization_mode = "skip_rematerialization"
```
obs-gh-abhinavpappu, line 132: remove the corresponding check below as well

observe/resource_dataset_test.go
```diff
@@ -527,6 +526,131 @@ func TestAccObserveDatasetErrors(t *testing.T) {
 	})
 }
 
+// Test edit-forward works when change is compatible
+func TestAccObserveDatasetEditForward(t *testing.T) {
```
obs-gh-abhinavpappu, line 530: thanks for the tests! this looks amazing

obs-gh-abhinavpappu (commented):
Looks great overall! The rbac tests failing is due to an API-side issue that should be fixed soon. `TestAccObserveDatasetUpdate` is also failing, left a comment to remove the now unnecessary check in that test.

obs-gh-abhinavpappu:
> Now I make an update to the app, and am making changes incompatible with edit_forward. Is my app upgrade hosed?

Hmm that's good point. I think with the current PR, if you deliberately make changes incompatible with edit_forward, you should also change the `rematerialization_mode` in the same commit (and then change it back afterwards, which isn't ideal, but terraform doesn't really give us a better way to do this). In which case, maybe it's actually be desirable to fail an app upgrade if it results in an unexpected rematerialization (and hopefully this would be caught by our own tests before reaching customers 😅)

Another possibility might be to have two separate rematerialization mode options, one that's like "always skip rematerialization", and another that's "skip rematerialization if possible"

obs-gh-jamesweng:
@danielodievich Good point, would an extra option emulating the current behavior work?

danielodievich:
> @danielodievich Good point, would an extra option emulating the current behavior work?

Yes, I think it is essential to introduce that option!

I think it is asking too much of someone authoring the app to add then remove then add then remove the property otherwise. The apps allow installation of any version backwards and then forwards! It would get very confusing and potentially introduce completely broken install in customer who leapfrogged a version or two..

client/meta/helpers.go
```diff
@@ -174,7 +174,8 @@ var AllAccelerationDisabledSource = []AccelerationDisabledSource{
 
 var AllRematerializationModes = []RematerializationMode{
 	RematerializationModeRematerialize,
-	RematerializationModeSkiprematerialization,
+	RematerializationModeSkipRematerialization,
```
obs-gh-abhinavpappu, line 177: nit: it's a bit confusing combining the GQL enum and the terraform-only option, i would rather keep everything in the `client/meta` package to be based only on the GQL, and then define a new enum in `observe/resource_dataset.go` that represents all the options available for the terraform field, and have them map over to the GQL ones, but not a huge deal

obs-gh-abhinavpappu (approved):

obs-gh-abhinavpappu (approved):

PR #209, opened 2025-03-05
fix: keep original behavior for skip_rematerialization, introduce new must_skip_rematerialization

obs-gh-abhinavpappu:
Changing the behavior of "skip_rematerialization" to fail if skipping rematerialization is not possible may cause issues with apps in certain rare cases. Changed it back to the original behavior, and introduced new option "must_skip_rematerialization" for new behavior.


obs-gh-jamesweng (approved):
LGTM, thanks for doing this @obs-gh-abhinavpappu !

PR #210, opened 2025-03-12
Add reports resource/datasource in provider.

georgios-observeinc:
- **Fixes formatting of inlined TF code**
- **feat: add support for reporting queries**


observe/resource_report.go
```diff
@@ -0,0 +1,402 @@
...
+				Required:    true,
+				MinItems:    1,
+				MaxItems:    1,
+				Description: descriptions.Get("report", "schema", "schedule", "description"),
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"frequency": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: descriptions.Get("report", "schema", "schedule", "frequency"),
```
obs-gh-abhinavpappu, line 142: Could you add validation for this and some of the below? Would be good to catch things during the `plan` phase instead of `apply`. Can add `ValidateDiagFunc: validateStringInSlice([]string{"Daily", "Weekly", "Monthly"})`

We also appear to have a convention of snake-case for enums in the terraform definition, see `validateEnums`. Though this seems mainly intended for GQL enums, and there are helpers to convert between snake and pascal.

observe/resource_report.go
```diff
@@ -0,0 +1,402 @@
...
+		setResourceData("last_run_status", *report.LastRunStatus)
+	}
+	if report.LastRunError != nil {
+		setResourceData("last_run_error", *report.LastRunError)
+	}
+
+	return diags
+}
+
+func resourceDefinitionFromResourceData(data *schema.ResourceData) (req *rest.ReportsDefinition, diags diag.Diagnostics) {
```
obs-gh-abhinavpappu, line 283: include something specific about `report` in the function name (since everything's in the same package), maybe `reportFromResourceData`

observe/resource_report.go
```diff
@@ -0,0 +1,402 @@
...
+	req = &rest.ReportsDefinition{
+		Schedule:        rest.ReportsSchedule{},
+		EmailRecipients: []string{},
```
obs-gh-abhinavpappu, line 286: unnecessary? looks like you're setting both `req.Schedule` and `req.EmailRecipients` below

observe/resource_report.go
```diff
@@ -0,0 +1,402 @@
...
+
+	return append(diags, reportToResourceData(report, data)...)
+
+}
+
+func resourceReportRead(ctx context.Context, data *schema.ResourceData, meta interface{}) (diags diag.Diagnostics) {
+	client := meta.(*observe.Client)
+	result, err := client.GetReport(ctx, data.Id())
+	if err != nil {
+		if rest.HasStatusCode(err, 404) {
```
obs-gh-abhinavpappu, line 385: nit: `http.StatusNotFound` instead of 404

obs-gh-abhinavpappu (commented):
Looks great overall! Left a few comments, mostly nits. Thanks for some of the code cleanup as well

observe/resource_report.go
georgios-observeinc: We used to have upper-snake-case, but switched to Pascal case. This is what the monitors REST API is ising:
https://api.observeinc.com/v1/openapi#:~:text=components/schemas/StageQuery%27-,MonitorV2AlarmLevel,-%3A%0A%20%20%20%20%20%20type%3A%20string%0A%20%20%20%20%20%20enum

obs-gh-abhinavpappu (approved):
lgtm!

PR #216, opened 2025-03-24
feat: support no data rule for monitor v2 in terraform

obs-gh-andrewlee:
Adds only the necessary components to support the no data rule feature for the threshold type monitor.

### DIFF
--- a/observe/resource_monitor_v2.go
+++ b/observe/resource_monitor_v2.go
@@ -1288,26 +1288,27 @@ func newMonitorV2ComparisonInput(path string, data *schema.ResourceData) (compar
 
 func newMonitorV2ThresholdRuleInput(path string, data *schema.ResourceData) (threshold *gql.MonitorV2ThresholdRuleInput, diags diag.Diagnostics) {
 	// required
-	compareValues := []gql.MonitorV2ComparisonInput{}
-	for i := range data.Get(fmt.Sprintf("%scompare_values", path)).([]interface{}) {
-		comparisonInput, diags := newMonitorV2ComparisonInput(fmt.Sprintf("%scompare_values.%d.", path, i), data)
-		if diags.HasError() {
-			return threshold, diags
-		}
-		compareValues = append(compareValues, *comparisonInput)
-	}
-
 	valueColumnName := data.Get(fmt.Sprintf("%svalue_column_name", path)).(string)
 	aggregation := gql.MonitorV2ValueAggregation(toCamel(data.Get(fmt.Sprintf("%saggregation", path)).(string)))
 
 	// instantiation
 	threshold = &gql.MonitorV2ThresholdRuleInput{
-		CompareValues:   compareValues,
 		ValueColumnName: valueColumnName,
 		Aggregation:     aggregation,
 	}
 
 	// optionals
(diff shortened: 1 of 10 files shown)


obs-gh-abhinavpappu:
FYI I'm taking a look at the failing Link-related tests, seems to have been failing for a couple days now.

obs-gh-andrewlee:
> FYI I'm taking a look at the failing Link-related tests, seems to have been failing for a couple days now.

Please lmk when it's fixed :)

observe/resource_monitor_v2.go
```diff
@@ -105,6 +105,55 @@ func resourceMonitorV2() *schema.Resource {
...
+				Description: descriptions.Get("monitorv2", "schema", "no_data_rules", "description"),
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"expiration": { // Duration
+							Type:             schema.TypeString,
+							Optional:         true,
+							ValidateDiagFunc: validateTimeDuration,
+							Description:      descriptions.Get("monitorv2", "schema", "no_data_rules", "expiration"),
+						},
+						"threshold": { // MonitorV2ThresholdRuleInput
```
obs-gh-abhinavpappu, line 121: If the structure of this is always going to match the `threshold` in `rules` (which it looks like it will since the GQL type is the same), would be good to move it into a common function. Something like
```
"threshold": schemaMonitorV2Threshold("no_data_rules"),
```

```
func schemaMonitorV2Threshold(descriptionBase string) *schema.Schema {
    return &schema.Schema {
        ...
        Description: descriptions.Get("monitorv2", "schema", descriptionBase, "threshold", "description")
    }
}
```

obs-gh-abhinavpappu (commented):
Looks great overall! And I believe the (server-side) change causing the link tests to fail has been turned off, so next run should hopefully pass

observe/resource_monitor_v2.go
obs-gh-andrewlee: I think this can be done in a separate PR instead of here.

observe/resource_monitor_v2.go
```diff
@@ -105,6 +105,55 @@ func resourceMonitorV2() *schema.Resource {
...
+				Optional:    true,
+				MaxItems:    1,
+				Description: descriptions.Get("monitorv2", "schema", "no_data_rules", "description"),
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"expiration": { // Duration
+							Type:             schema.TypeString,
+							Optional:         true,
+							ValidateDiagFunc: validateTimeDuration,
+							Description:      descriptions.Get("monitorv2", "schema", "no_data_rules", "expiration"),
```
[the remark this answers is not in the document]
obs-gh-andrewlee: Done :)

observe/resource_monitor_v2.go
obs-gh-abhinavpappu: ack

obs-gh-abhinavpappu (approved):
lgtm! double checked the export flow as well and works great

obs-gh-abhinavpappu:
the report test is non-blocking, and the failures in there look fine (there was a failure due to timeout and it retried, which then succeeded)

feel free to merge whenever you want to!

PR #222, opened 2025-04-08
test: update static resource names in tests to random strings

obs-gh-abhinavpappu:
necessary to avoid conflicts when tests are running simultaneously, as well as indicate to the sweeper what to delete


vikramraman (approved):
lgtm. May want to look into the failing tests though (unsure if it's tied to the issue this aims to address)

PR #230, opened 2025-05-06
feat: add ReportManager grant role

obs-gh-abhinavpappu:


client/internal/meta/schema/types.generated.graphql
```diff
@@ -87,6 +87,7 @@ enum RbacRole @goModel(model: "observe/access/permission.Role") {
     Lister
     Manager
     MonitorGlobalMute
+    ReportManager
```
vikramraman, line 90: Should we add the `ReportManage` action under AccessAction for RBAC v2 as well?

obs-gh-abhinavpappu: yeah i'll go ahead and add it, though we don't actually use the `AccessAction`s in the terraform provider

vikramraman (approved):

PR #231, opened 2025-05-07
feat: Add scheduled monitor stuff

obs-gh-catherman:
This adds both the intervals and the cron stuff.

### DIFF
--- a/docs/data-sources/monitor_v2.md
+++ b/docs/data-sources/monitor_v2.md
@@ -44,7 +44,7 @@ data "observe_monitor_v2" "name_lookup" {
 One of `name` or `id` must be set.
 - `name` (String) Monitor name.
 One of `name` or `id` must be set. If `name` is provided, `workspace` must be set.
-- `scheduling` (Block List) Holds information about when the monitor should evaluate. The types of scheduling (interval, transform) are exclusive. If ommitted, defaults to transform. (see [below for nested schema](#nestedblock--scheduling))
+- `scheduling` (Block List) Holds information about when the monitor should evaluate. The types of scheduling (interval, transform, and scheduled) are exclusive. If omitted, defaults to transform. (see [below for nested schema](#nestedblock--scheduling))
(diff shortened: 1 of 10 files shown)


docs/data-sources/monitor_v2.md
```diff
@@ -44,6 +44,7 @@ data "observe_monitor_v2" "name_lookup" {
 One of `name` or `id` must be set.
 - `name` (String) Monitor name.
 One of `name` or `id` must be set. If `name` is provided, `workspace` must be set.
+- `scheduling` (Block List) Holds information about when the monitor should evaluate. The types of scheduling (interval, transform) are exclusive. If ommitted, defaults to transform. (see [below for nested schema](#nestedblock--scheduling))
```
obs-gh-abhinavpappu, line 47: update the "(interval, transform) are exclusive" to include `scheduled` in `descriptions/monitorv2.yaml`

also you didn't add this but could you fix "ommitted" to be "omitted"

observe/resource_monitor_v2.go
```diff
@@ -292,6 +292,51 @@ func resourceMonitorV2() *schema.Resource {
 								},
 							},
 						},
+						"interval": { // MonitorV2IntervalScheduleInput
```
obs-gh-abhinavpappu, line 295: could add a `ConflictsWith: []string{"scheduling.0.transform", "scheduling.0.scheduled"}` here and similar for the other two

obs-gh-catherman: In the past, I had trouble getting this to work right with nested objects. I'll try again.

obs-gh-catherman:
I think we need to merge soon because intervals are in the wild. Cronitors won't really get exercised until we tell people to use it, so I feel ok getting this going. What do you think?

obs-gh-abhinavpappu:
I'll update this PR adding a note saying support for cronitors is still experimental and merge it in, thanks @obs-gh-catherman !

PR #232, opened 2025-05-08
feat: support generating bindings for monitors

obs-gh-abhinavpappu:
Similar to the existing approach for dashboards, generates "bindings" to a terraform local variable for each id found in the monitor definition to support cross-tenant exports. Unlike with dashboards, monitors don't have a `layout` field to insert these bindings into for later consumption, so instead added a new `_bindings` field to the data source to hold this information.

Broken into separate commits for ease of reviewing.

### DIFF
--- a/client/binding/binding.go
+++ b/client/binding/binding.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	observe "github.com/observeinc/terraform-provider-observe/client"
-	"github.com/observeinc/terraform-provider-observe/client/meta/types"
 	"github.com/observeinc/terraform-provider-observe/client/oid"
 )
 
@@ -27,11 +26,12 @@ type ResourceCache struct {
 	idToLabel       map[Ref]ResourceCacheEntry
 	workspaceOid    *oid.OID
 	workspaceEntry  *ResourceCacheEntry
-	forResourceKind string
+	forResourceKind Kind
 	forResourceName string
 }
 
-func NewResourceCache(ctx context.Context, kinds KindSet, client *observe.Client, forResourceKind string, forResourceName string) (ResourceCache, error) {
+// NewResourceCache loads all resources of the given kinds and creates a cache of id -> label mappings
+func NewResourceCache(ctx context.Context, kinds KindSet, client *observe.Client, forResourceKind Kind, forResourceName string) (ResourceCache, error) {
 	var cache = ResourceCache{
 		idToLabel:       make(map[Ref]ResourceCacheEntry),
 		forResourceKind: forResourceKind,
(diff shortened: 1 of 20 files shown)


observe/data_source_dashboard.go
```diff
@@ -75,12 +76,59 @@ func dataSourceDashboardRead(ctx context.Context, data *schema.ResourceData, met
...
-	data.SetId(ws.Id)
+	data.SetId(dashboard.Id)
 
-	return dashboardToResourceData(ctx, ws, data, client, true)
+	diags = dashboardToResourceData(dashboard, data)
+	if diags.HasError() {
+		return diags
+	}
+
+	if client.ExportObjectBindings {
```
vikramraman, line 91: nit: Could this be moved into a helper func (within this file) that clearly documents the use case / functionality.

observe/data_source_dashboard.go
```diff
@@ -75,12 +76,59 @@ func dataSourceDashboardRead(ctx context.Context, data *schema.ResourceData, met
...
+		if err != nil {
+			return diag.Errorf("Failed to initialize binding generator: %s", err.Error())
+		}
+
+		// generate binding for workspace
+		if err := data.Set("workspace", gen.TryBind(binding.KindWorkspace, dashboard.WorkspaceId)); err != nil {
+			return diag.FromErr(err)
+		}
+
+		// generate bindings for workspace, stages, parameters, parameter_values, and layout,
```
vikramraman, line 103: nit: can remove `workspace` from this comment as it is handled just above

observe/data_source_monitor_v2.go
```diff
@@ -542,5 +549,44 @@ func dataSourceMonitorV2Read(ctx context.Context, data *schema.ResourceData, met
 	}
 
 	data.SetId(m.Id)
-	return resourceMonitorV2Read(ctx, data, meta)
+	diags = monitorV2ToResourceData(ctx, m, data, client)
+	if diags.HasError() {
+		return diags
+	}
+
+	if client.ExportObjectBindings {
```
vikramraman, line 557: Similar feedback to move this into a helper func in this file

client/binding/binding.go
```diff
@@ -75,6 +75,14 @@ func NewResourceCache(ctx context.Context, kinds KindSet, client *observe.Client
 			for _, user := range users {
 				cache.addEntry(KindUser, user.Label, user.Id.String(), true, &disambiguator, existingResourceNames)
 			}
+		case KindMonitorV2Action:
```
vikramraman, line 78: Curious, do we need to support monitor actions amongst the bindings part?

If it's not needed, it'd be nice to revert the change to the search api to return a list of actions as well.

vikramraman (changes requested):

client/binding/binding.go
obs-gh-abhinavpappu: Yeah we do need to support monitor actions for cross-tenant export of any monitors that include shared actions. Otherwise they would be left as ids in the export

vikramraman (approved):

PR #234, opened 2025-05-15
feat: terraform resource for ingest filters

obs-gh-yihongliu:


client/meta/types/user_id.go
```diff
@@ -19,7 +19,7 @@ func (u *UserIdScalar) UnmarshalJSON(b []byte) error {
 		return fmt.Errorf("could not parse userId as JSON string: %w", err)
 	}
 	uid, err := strconv.ParseInt(s, 10, 64)
-	if err != nil || uid < 1 {
```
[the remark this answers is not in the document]
obs-gh-abhinavpappu: Yeah I think that's fine, we shouldn't be validating this during Unmarshal anyway.

observe/resource_ingest_filter.go
```diff
@@ -0,0 +1,191 @@
...
+	data.SetId(filter.Id)
+
+	return append(diags, resourceIngestFilterRead(ctx, data, meta)...)
+}
+
+func resourceIngestFilterRead(ctx context.Context, data *schema.ResourceData, meta interface{}) (diags diag.Diagnostics) {
+	client := meta.(*observe.Client)
+	result, err := client.GetIngestFilter(ctx, data.Id())
+	if err != nil {
+		if gql.HasErrorCode(err, gql.ErrNotFound) {
```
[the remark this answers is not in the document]
obs-gh-yihongliu: Confirmed from api - it is just generated with gmodelgen.

```
{
  "errors": [
    {
      "message": "No IngestFilter with id 100101 is available.",
      "path": [
        "ingestFilter"
      ],
      "extensions": {
        "code": "NOT_FOUND"
      }
    }
  ],
  "data": null
}
```

observe/resource_ingest_filter.go
```diff
@@ -0,0 +1,191 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: descriptions.Get("ingest_filter", "schema", "name"),
+			},
+			"drop_rate": {
+				Type:        schema.TypeFloat,
+				Optional:    true,
+				Description: descriptions.Get("ingest_filter", "schema", "drop_rate"),
+			},
+			"enabled": {
```
obs-gh-abhinavpappu, line 54: probably want this to default to `true`

observe/resource_ingest_filter.go
```diff
@@ -0,0 +1,191 @@
...
+			},
+		},
+	}
+}
+
+func newIngestFilterConfig(data *schema.ResourceData) (input *gql.IngestFilterInput, diags diag.Diagnostics) {
+	var (
+		name               = data.Get("name").(string)
+		pipeline           = data.Get("pipeline").(string)
+		dropRate           = data.Get("drop_rate").(float64)
```
obs-gh-abhinavpappu, line 73: looks like `drop_rate` is optional, so this will just give you the zero value of `0.0` if it's unspecified, is that what you want?

obs-gh-yihongliu: we want to default to 1.0.

obs-gh-abhinavpappu: Got it, in that case, you might want to specify the default as `1.0`. But actually assuming this is an optional field in the GQL API as well, you can use `.GetOk` instead to detect if the field is set by the user or not, and if not, can leave it undefined in the GQL input as well, and let the API decide what to default to.

observe/resource_ingest_filter.go
```diff
@@ -0,0 +1,191 @@
+package observe
```
obs-gh-abhinavpappu, line 1: Can we rename this file to `resource_drop_filter.go`, generally want these to match the terraform resource name. Same for `resource_ingest_filter_test.go` and `descriptions/ingest_filter.yaml`. Can leave the files in `meta` as `ingest_filter` though since that's what the graphql type is called.

observe/resource_ingest_filter.go
obs-gh-yihongliu: hmm actually I think I got mixed up a bit here. From the observe frontend it defaults to 1.0, but in our graphql it is a required field. I will change this to required.

observe/resource_ingest_filter.go
obs-gh-yihongliu: ah from an api perspective this is also a required field, I will change it.

obs-gh-abhinavpappu (commented):
Lastly could you add an example at `examples/resources/observe_drop_filter/resource.tf` (can probably just copy your example in the test case, look at the example for `monitor_v2` as a reference). Also add an `import.sh` in that folder with the content `terraform import observe_drop_filter.example 1234567`

But otherwise looks great overall!

observe/resource_ingest_filter.go
obs-gh-abhinavpappu: Sounds good. It also may be reasonable to give the field a default of `1.0`, but I'll leave it up to you

observe/resource_ingest_filter.go
obs-gh-abhinavpappu: actually for this one, i think it would make more sense to leave it as optional and have it default to `true` since if you just have
```
resource "observe_drop_filter" "example" {
  workspace      = data.observe_workspace.default.oid
  name           = "test-filter"
  pipeline       = "filter FIELDS.x ~ y"
  source_dataset = data.observe_datastream.example.dataset
  drop_rate      = 0.99
}
```
defined in the terraform, it's pretty reasonable to assume that this drop filter is enabled, and having an explicit `enabled = true` is mostly redundant

for `drop_rate`, i think there's some value in requiring an explicit `drop_rate = 1.0` in the terraform for clarity, but i can see it either way

observe/resource_drop_filter_test.go
```diff
@@ -0,0 +1,91 @@
...
+					enabled = true
+				}`, randomPrefix),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("observe_drop_filter.example", "name", randomPrefix+"-filter"),
+					resource.TestCheckResourceAttr("observe_drop_filter.example", "drop_rate", "0.99"),
+					resource.TestCheckResourceAttr("observe_drop_filter.example", "pipeline", "filter FIELDS.x ~ y"),
+					resource.TestCheckResourceAttr("observe_drop_filter.example", "enabled", "true"),
+					resource.TestCheckResourceAttrPair("observe_drop_filter.example", "source_dataset", "observe_datastream.test", "dataset"),
+				),
+			},
```
obs-gh-abhinavpappu, line 88: Could you also check that updating a resource works properly. Can do so by adding another test step and copying the `Config` from the above step over, updating each of the fields (except workspace).

The test works by modifying the same terraform configuration file for each step and preserving the state. So by including the same resource in the 2nd step, it'll be updating the existing resource created in the first step, rather than creating a new resource.

obs-gh-abhinavpappu (approved):
LGTM!

obs-gh-abhinavpappu:
ah you need to run `make generate` again after the Optional change

PR #237, opened 2025-05-20
fix: ingest-filter enabled field not accounted for

obs-gh-yihongliu:
I noticed that if I disabled the filter from ui, tf wouldn't see the difference. This should fix the issue.


obs-gh-abhinavpappu (approved):
ah good catch

PR #240, opened 2025-05-23
fix: add example for drop filters

obs-gh-yihongliu:


obs-gh-yihongliu:
oops this is not the right way to add the example

PR #243, opened 2025-06-05
feat: terraform resource for ingest tokens

obs-gh-davidzhang:
Allows user to manage ingest tokens via Terraform.

Explicitly does not support hardcoded routing so the `updateIngestTokenAssociation` API is unused, see https://observe.atlassian.net/browse/OB-42425 for details. Users will have to do routing using a package header, `X-Observe-Target-Package`, although this is usually abstracted away in observe-agent.


observe/resource_ingest_token.go
```diff
@@ -0,0 +1,209 @@
...
+		// authtokens, which should never apply here.
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  fmt.Sprintf("failed to create ingest token [id=%s]", data.Id()),
+			Detail:   "failed to create secret",
+		})
+		return diags
+	}
+	data.SetId(ingestToken.Id)
+	if err := data.Set("secret", ingestToken.Secret); err != nil {
```
obs-gh-abhinavpappu, line 151: i'm assuming this secret is only returned as part of the create response and not in the read? could you add a comment mentioning that here

client/internal/meta/operation/ingest_token.graphql
```diff
@@ -0,0 +1,49 @@
+fragment IngestToken on IngestToken {
+  id
+  name
+  description
+  disabled
+  secret
+}
+
+mutation createIngestToken($workspaceId: ObjectId!, $name: String, $description: String) {
```
obs-gh-abhinavpappu, line 9: Any reason to not have this accept an `$input: IngestTokenInput` like the update query? Then could have `newIngestTokenResource` return `gql.IngestTokenInput` directly and get rid of `IngestTokenResource`. I see you have `disabled: false` hardcoded, but looks like you're already ensuring disabled is false in `resourceIngestTokenCreate`.

observe/resource_ingest_token_test.go
```diff
@@ -0,0 +1,100 @@
...
+			Providers: testAccProviders,
+			Steps: []resource.TestStep{
+				{
+					Config: configPreamble + `
+					resource "observe_ingest_token" "example" {
+						workspace = data.observe_workspace.default.oid
+						name = "name-0"
+						description = "this is a description"
+						disabled = false
+					}
```
obs-gh-abhinavpappu, line 49: Please add an example at `examples/resources/observe_ingest_token/resource.tf`, can copy what you have in the test. See https://github.com/observeinc/terraform-provider-observe/blob/master/examples/resources/observe_drop_filter/resource.tf

observe/resource_ingest_token.go
```diff
@@ -0,0 +1,209 @@
...
+)
+
+func resourceIngestToken() *schema.Resource {
+	return &schema.Resource{
+		Description:   descriptions.Get("ingest_token", "description"),
+		CreateContext: resourceIngestTokenCreate,
+		ReadContext:   resourceIngestTokenRead,
+		UpdateContext: resourceIngestTokenUpdate,
+		DeleteContext: resourceIngestTokenDelete,
+		Importer: &schema.ResourceImporter{
```
obs-gh-abhinavpappu, line 22: Hmm to properly support importing an existing ingest token into terraform, we would probably need a custom importer that also requires the user to provide the secret so we can set it in the state (since we can't fetch it using just the id).

Looks like for datastream tokens, there's a custom importer that looks for an environment variable called SECRET and uses that. I would probably do something similar, see this PR: https://github.com/observeinc/terraform-provider-observe/pull/28/files for the necessary changes. Also fine if you want to punt on it until the need arises.

obs-gh-abhinavpappu (commented):
looks great overall! left some nits

obs-gh-abhinavpappu:
Also FYI, the failing `report` test is fine and does not block merging. The `acceptance` tests will retry a few times on failure since there may be things outside of our control causing a test to fail like network issues or a server restart. So the `report` test is just there to inform us of any tests that failed in one of the runs (but a retry passed) in case there's an actual problem. And all the issues reported for this PR are known and not concerning.

observe/resource_ingest_token.go
```diff
@@ -0,0 +1,209 @@
...
+			"oid": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: descriptions.Get("common", "schema", "oid"),
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: descriptions.Get("ingest_token", "schema", "name"),
```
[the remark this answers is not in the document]
obs-gh-davidzhang: We don't put any restrictions on the name, so I think this is OK.

observe/resource_ingest_token.go
```diff
@@ -0,0 +1,209 @@
...
+			},
+			"oid": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: descriptions.Get("common", "schema", "oid"),
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
```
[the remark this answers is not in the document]
obs-gh-davidzhang: Yes. It will be autogenerated like `ingest-token-[timestamp]`, we have similar behavior for filedrop atm.

observe/resource_ingest_token.go
obs-gh-davidzhang: Done, changed the API to also return workspaceId to make this possible.

client/internal/meta/operation/ingest_token.graphql
obs-gh-davidzhang: Changed it to do this.

observe/resource_ingest_token_test.go
obs-gh-davidzhang: Done.

obs-gh-abhinavpappu (approved):

observe/resource_ingest_token.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	client := meta.(*observe.Client)
+	workspaceId := data.Get("workspace").(string)
+	resource := ingestTokenInputFromResource(data)
+	if resource.Disabled != nil && *resource.Disabled {
+		return handleError("?", "ingest token cannot be disabled on creation")
+	}
+
+	ingestToken, err := client.CreateIngestToken(ctx, workspaceId, gql.IngestTokenInput{
```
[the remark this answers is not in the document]
obs-gh-davidzhang: Done

obs-gh-abhinavpappu (approved):

PR #253, opened 2025-07-11
test: move TestAccObserveReport out of unit test

chenjiayuan:


obs-gh-abhinavpappu (approved):
The test failure seems to be due to an API-side change in this environment. Will hopefully be fixed soon after which we can re-run the tests, but otherwise can skip `TestAccObserveSourceDatasetResource` if not

chenjiayuan:
> The test failure seems to be due to an API-side change in this environment. Will hopefully be fixed soon after which we can re-run the tests, but otherwise can skip `TestAccObserveSourceDatasetResource` if not

do you know anyone current working on the fix?

PR #260, opened 2025-07-23
tests: Fix dataset edit-forward tests

obs-gh-jamesweng:
Since dataset schemas can be changed now, change the edit-forward test to use make_resource instead - changing the primary key set should always trigger rematerialization.


obs-gh-abhinavpappu (approved):
lgtm!

PR #264, opened 2025-08-19
feat: resource for service accounts

vikramraman:


client/oid/oid.go
```diff
@@ -55,6 +55,7 @@ const (
 	TypeReport                  Type = "report"
 	TypeIngestFilter            Type = "ingestfilter"
 	TypeIngestToken             Type = "ingesttoken"
+	TypeServiceAccount          Type = "serviceaccount"
```
vikramraman, line 58: I'm not sure whether we want to just reuse the `user` oid instead?

That would make integrating with the rbac resources (group_member, grant etc) seamless.

vikramraman: Looked into the RBAC aspects. I see a few options:

1. The service account resource returns a user oid as it's just another kind of user.
2. The service account resource stays as is (returns a service account oid, and an id), and one can construct a user oid out of it:
```
   data "observe_oid" "sa_oid" {
     id = observe_service_account.example.id
     type = "user"
   }
```
3. The service account resource returns two fields: `oid` of type service account and a `user_oid` of type user.

option 1 offers the best end user ergonomics, and makes it very clear service accounts are just special kind of users.

PR #268, opened 2025-09-08
OB-49319: Document interval monitor deprecation

obs-gh-christysong:
This is a small change to the documentation informing of the deprecation of the `interval` type monitors.

### DIFF
--- a/observe/descriptions/monitorv2.yaml
+++ b/observe/descriptions/monitorv2.yaml
@@ -63,7 +63,7 @@ schema:
       deprecation: |
          Interval scheduling is deprecated. Creation of new interval monitors is no longer supported. Use transform or scheduled monitor instead.
       description: |
-        @DEPRECATED: Creation of new interval monitors is not supported, but existing interval monitors will continue to be supported. 
+        Creation of new interval monitors is not supported, but existing interval monitors will continue to be supported. 
         Recommended to migrate to transform scheduling if the pre-existing interval monitor runs on an accelerable OPAL query.
         Was used to run explicit ad-hoc queries.
       interval: |
--- a/observe/descriptions/monitorv2.yaml
+++ b/observe/descriptions/monitorv2.yaml
@@ -58,10 +58,14 @@ schema:
     Describes the groups that logically separate events/rows/etc from each other. If monitor dataset is resource type and monitor strategy is promote, this field should be either empty or only contain the primary keys of the dataset.
   scheduling:
     description: |
-      Holds information about when the monitor should evaluate. The types of scheduling (interval, transform, and scheduled) are exclusive. If omitted, defaults to transform.
+      Holds information about when the monitor should evaluate. The types of scheduling (transform, scheduled, and interval@deprecated) are exclusive. If omitted, defaults to transform.
     interval:
+      deprecation: |
+         Interval scheduling is deprecated. Creation of new interval monitors is no longer supported. Use transform or scheduled monitor instead.
       description: |
-        Should be used to run explicit ad-hoc queries.
+        @DEPRECATED: Creation of new interval monitors is not supported, but existing interval monitors will continue to be supported. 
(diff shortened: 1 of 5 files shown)


observe/descriptions/monitorv2.yaml
```diff
@@ -58,10 +58,14 @@ schema:
...
   scheduling:
     description: |
-      Holds information about when the monitor should evaluate. The types of scheduling (interval, transform, and scheduled) are exclusive. If omitted, defaults to transform.
+      Holds information about when the monitor should evaluate. The types of scheduling (transform, scheduled, and interval@deprecated) are exclusive. If omitted, defaults to transform.
     interval:
+      deprecation: |
+         Interval scheduling is deprecated. Creation of new interval monitors is no longer supported. Use transform or scheduled monitor instead.
       description: |
-        Should be used to run explicit ad-hoc queries.
+        @DEPRECATED: Creation of new interval monitors is not supported, but existing interval monitors will continue to be supported.
```
obs-gh-abhinavpappu, line 66: nit: Don't need to add the `@DEPRECATED:` here. If you look at the generated markdown, it already has "interval (Block List, Max: 1, **Deprecated**) @‎‎​DEPRECATED: ..." (comes from marking it deprecated in `resource_monitor_v2.go`)

observe/descriptions/monitorv2.yaml
```diff
@@ -58,10 +58,14 @@ schema:
     Describes the groups that logically separate events/rows/etc from each other. If monitor dataset is resource type and monitor strategy is promote, this field should be either empty or only contain the primary keys of the dataset.
   scheduling:
     description: |
-      Holds information about when the monitor should evaluate. The types of scheduling (interval, transform, and scheduled) are exclusive. If omitted, defaults to transform.
+      Holds information about when the monitor should evaluate. The types of scheduling (transform, scheduled, and interval@deprecated) are exclusive. If omitted, defaults to transform.
```
obs-gh-abhinavpappu, line 61: nit: i think `interval@deprecated` looks a bit weird, maybe just get rid of `interval` altogether, but also fine as is, should be clear enough

obs-gh-abhinavpappu (approved):
lgtm!

observe/descriptions/monitorv2.yaml
obs-gh-christysong: I thought it would be best not to remove it completely so that people aren't confused when they encounter an interval monitor but don't see it in some parts of the doc since we are still allowing for pre-existing ones. But, we would still want to mark it as deprecated so to show we don't support new ones any more. What do you think? If the "@deprecated" is a bit of a weird style, do you think there's a better way to mark this?

observe/descriptions/monitorv2.yaml
obs-gh-abhinavpappu: i think if you add `Deprecated: something` to `data_source_monitor_v2.go` as well [here](https://github.com/observeinc/terraform-provider-observe/blob/d612630b35cac48d6fadef19398d61bfe84f7c11/observe/data_source_monitor_v2.go#L290), the `Deprecated` tag will show up

but also it's not as important to add it to the data source since the field is read-only (actually not quite fully true due to some hacks that were necessary to get exporting to work nicely, but effectively true); the data source is only used for reading data about an existing monitor, not creating/updating monitors

observe/descriptions/monitorv2.yaml
obs-gh-abhinavpappu: something like strikethrough would be cool but i don't think that's possible with how the docs are generated from the yaml file, `@deprecated` is fine

observe/descriptions/monitorv2.yaml
obs-gh-christysong: Gotcha! Just made this small tweak! I have the proper "(Block List, Deprecated)" now. Removed the all caps tag.

PR #310, opened 2026-04-13
feat: add OAuth2 client credentials (M2M) authentication

obs-gh-jesseswidler:
Add an `oauth2` provider block that performs the OAuth2 client_credentials grant to obtain a JWT from an external IdP (e.g. Microsoft Entra ID), then uses it as the bearer token for all API requests.


observe/provider.go
```diff
@@ -257,6 +295,22 @@ func getConfigureContextFunc(userAgent func() string) schema.ConfigureContextFun
...
+			oauth2List := v.([]interface{})
+			if len(oauth2List) > 0 {
+				oauth2Map := oauth2List[0].(map[string]interface{})
+				oc := &observe.OAuth2Config{
+					ClientID:     oauth2Map["client_id"].(string),
+					ClientSecret: oauth2Map["client_secret"].(string),
+					TokenURL:     oauth2Map["token_url"].(string),
+				}
+				if scope, ok := oauth2Map["scope"].(string); ok && scope != "" {
+					oc.Scopes = []string{scope}
```
obs-gh-abhinavpappu, line 308: is/will there potentially be a use case for supporting multiple scopes in the provider config?

obs-gh-abhinavpappu (approved):

observe/provider.go
obs-gh-jesseswidler: We don't need it currently - you would create different service accounts with varying levels of permissions depending on the needs of the machine driven workload.

But potentially it would be useful.  For instance - if you wanted to create "on demand" users from the IDP, then the scope could be used to say what group(s) they should go in.  This is how [Snowflake uses it in step 2.9 in this doc.](https://community.snowflake.com/s/article/Create-External-OAuth-Token-Using-Azure-AD-For-The-OAuth-Client-Itself).  Another use case might be to apply a read-only scope to the requests.

PR #311, opened 2026-04-17
Fixes202604

georgios-observeinc:


GNUmakefile
```diff
@@ -6,7 +6,7 @@ VERSION?=$(shell git describe --tags --always)
 TESTARGS?=
 SWEEP?=^tf-\\d{16}
 SWEEP_DIR?=./observe
-OBSERVE_ROOT?=../observe
+OBSERVE_ROOT?=/home/dev/observe
```
obs-gh-abhinavpappu: could you make this `~/observe` instead (my main development environment has the folder at `/home/{user}/observe`), or perhaps add support for looking in multiple locations for the `observe` directory

GNUmakefile
```diff
@@ -77,10 +77,10 @@ copy-gql-schema:
 	find client/internal/meta/schema -name '*.graphql' -print0 | xargs -0 perl -0777 -i -pe 's/(\n +"""\n(.+\n)+? +""")?\n.+(?<!directive )\@eol.*//g'
 
 	# Strip out @owner directives from all .graphql files
-	find client/internal/meta/schema -name '*.graphql' -print0 | xargs -0 sed -i '' '/^directive/!s/ @owner([^)]*)//g'
+	find client/internal/meta/schema -name '*.graphql' -print0 | xargs -0 sed -i'' -e '/^directive/!s/ @owner([^)]*)//g'
```
obs-gh-abhinavpappu, line 82: ah i originally wrote this locally on my mac and looks like this is a difference between mac and linux

ideally would be nice to support both if possible, but regardless this change is fine since the main developer environment is linux (and i've since switched to developing terraform provider from there as well)

observe/data_source_app_version_test.go
obs-gh-abhinavpappu, line 31: the changes below no longer test a prerelease version as the test name implies

and are you sure the changes here are necessary to get the test to pass? the version it's querying does still exist: 
Image

observe/data_source_rbac_group_test.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	defaultRbacGroupReaderName = "reader"
+	defaultRbacGroupReaderName = "Everyone"
```
obs-gh-abhinavpappu, line 11: was there a test failure due to this? probably a good idea to change it to `Everyone` regardless, but the `reader` group does indeed still exist in the tenant used for CI tests so I wouldn't expect test failures

observe/data_source_app_version_test.go
georgios-observeinc: I have no idea why, but it does not exist on a testbox (after copying the flags from eng)
image

observe/data_source_rbac_group_test.go
georgios-observeinc: Same as above. In testbox, there is no `reader` group :|

image

observe/data_source_app_version_test.go
obs-gh-abhinavpappu: oh there's a `REGISTRY_INCLUDE_PRERELEASES` environment variable, not sure why we don't set it to `true` in testbox

but since it seems like the intent of this test is to test the prerelease versions, let's just skip this test in non-CI environments instead of changing the test, similar to what's done here: https://github.com/observeinc/terraform-provider-observe/blob/90ed7076d7398b85a59cce8fb2b377155fe66e75/observe/resource_filedrop_test.go#L16

observe/data_source_rbac_group_test.go
obs-gh-abhinavpappu: yep that's expected, all new tenants will not have the `reader` group by default, just tenants that have migrated over from rbac v1

i initially thought the tests failed in CI which would have been concerning

but yes i agree we should change it to `Everyone` (or `admin`) for robustness as those groups are guaranteed to exist in all tenants

GNUmakefile
georgios-observeinc: Sure, that works too. 

I thought that our scripts set up `/home/dev` to be a symlink to `/home/${USER}`

observe/data_source_app_version_test.go
georgios-observeinc: Ack!

observe/data_source_rbac_group_test.go
georgios-observeinc: Ack!

obs-gh-abhinavpappu (approved):

PR #314, opened 2026-04-29
Fix log-derived metric dataset GraphQL mutations

obs-gh-paritoshgarg:
## Summary
Fixes log-derived metric dataset acceptance tests by removing redundant `query` parameter from GraphQL mutations. Builds on PR #304 by @obs-gh-georgemanos.

## Changes
- Removed `$query` parameter from `saveLogDerivedMetricDataset` and `saveLogDerivedMetricDatasetDryRun` mutations
- The shaping query is already included in `logDerivedMetricQuery.shapingQuery`
- Regenerated GraphQL client code
- Updated test to validate API constraint (aggregation function cannot be changed)


client/meta/dataset.go
```diff
@@ -114,7 +114,7 @@ func (d *Dataset) Oid() *oid.OID {
 
 // SaveLogDerivedMetricDataset creates or updates a log-derived metric dataset.
 func (client *Client) SaveLogDerivedMetricDataset(ctx context.Context, workspaceId string, input *DatasetInput, queryInput *MultiStageQueryInput, ldmInput *LogDerivedMetricDefinitionInput, dependencyHandling *DependencyHandlingInput) (*LogDerivedMetricDataset, error) {
```
obs-gh-abhinavpappu: get rid of `queryInput` here too

observe/resource_log_derived_metric_dataset.go
```diff
@@ -0,0 +1,496 @@
...
+		return gql.StageQueryInput{}, diag.Errorf("input: empty dataset id")
+	}
+	if _, err := strconv.ParseInt(datasetID, 10, 64); err != nil {
+		return gql.StageQueryInput{}, diag.FromErr(fmt.Errorf("input: %w", errObjectIDInvalid))
+	}
+
+	pipeline := data.Get("shaping_query").(string)
+	// if pipeline == "" {
+	// 	pipeline = "filter true"
+	// }
```
obs-gh-abhinavpappu: nit: remove

observe/resource_log_derived_metric_dataset_test.go
```diff
@@ -0,0 +1,414 @@
...
+					aggregation {
+						function = "count"
+					}
+				}`, randomPrefix),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("observe_log_derived_metric_dataset.test", "description", "initial description"),
+					resource.TestCheckResourceAttr("observe_log_derived_metric_dataset.test", "metric_name", "update_request_count"),
+					resource.TestCheckResourceAttr("observe_log_derived_metric_dataset.test", "aggregation.0.function", "count"),
+				),
+			},
```
obs-gh-abhinavpappu, line 166: could we also add a "good" update case

currently the first step here creates the `observe_log_derived_metric_dataset` resource, and the second step ensures that the aggregation function can't be updated, but would be nice to also include a step that verifies that all fields that should be updatable can in fact be updated correctly

obs-gh-abhinavpappu (approved):

client/meta/dataset.go
obs-gh-paritoshgarg: Ack

observe/resource_log_derived_metric_dataset.go
obs-gh-paritoshgarg: Done

observe/resource_log_derived_metric_dataset_test.go
obs-gh-paritoshgarg: Ack

PR #319, opened 2026-05-03
OB-56990: accept both flat and nested DatasetRef REST shapes

obs-gh-matthewso:
## Summary

- The Observe sharein API is migrating `DatasetRef` from flat `{id, label?}` to nested `{id, record?: DatasetBrief{label, description, icon}}`. This change adds a custom `UnmarshalJSON` on REST `DatasetRef` that transparently accepts BOTH wire shapes and normalises to the nested in-memory form, so the single consumer (`GetInboundShareTable`) is shape-agnostic and the provider release can ship before the API server rolls over for the affected customer.
- The flat `Label` Go field is removed; consumers read through `Record.Label`. Nested wins when both fields arrive simultaneously during a transitional rollout. Structural error messages in `GetInboundShareTable` (`sourceDataset`, `sourceDataset.id`, `sourceDataset.record`, `sourceDataset.record.label`) let operators distinguish a missing record from a record that lost its label.
- No Terraform attribute schema changes, no state migration; the GraphQL enrichment fallback in `observe/resource_inbound_share_table.go` is untouched.

Rollout: ship this provider → ping the customer to upgrade → API server rolls over. Post-migration, `UnmarshalJSON` can be deleted (a trivial follow-up); the default struct unmarshaller is sufficient for nested-only wire traffic.

## Test plan

- [ ] Review the decoder normalisation in `client/rest/inbound_share_types.go` — confirm "nested wins", flat synthesises a brief with just the label, id-only leaves `Record` nil, top-level JSON null decodes to the zero value via the default path.
- [ ] Review the four structural error branches in `GetInboundShareTable` (`client/rest/inbound_share_tables.go`).
- [ ] Confirm `go build ./...`, `go vet ./...`, and `go test ./client/rest/... -count=1` are clean.
- [ ] Unit coverage: `TestDatasetRef_UnmarshalJSON` is a single table test with eleven subtests covering nested, flat, id-only, both-present, empty-label, malformed, record-wrong-type, explicit-null on both `record` and `label`, non-object `record`, and duplicate-key last-wins semantics.
- [ ] End-to-end coverage: `TestTrackTableResponseMarshaling_NestedShape`, `TestGetInboundShareTable_NestedShape`, and `TestGetInboundShareTable_IdOnlyResponse` exercise the new wire shape through the HTTP → decoder → validator pipeline. `TestUpdateInboundShareTable`'s server fixture is deliberately kept on the flat shape to prove the flat→nested normalisation end-to-end.

## Notes

- Contains one incidental commit (`chore: ignore .worktrees/ directory`) adding `.worktrees/` to `.gitignore` — made during worktree-based development of this change and useful for future worktree work on this repo. Happy to split it out if preferred.
- The failing `report` check is a pre-existing repo-wide issue (`TestAccObserveLinkCreate` flakes against the shared Observe test stack with a `Client.Timeout` on `/v1/meta`; test retry succeeds but the xunit parser flags the one failed attempt). Every recent PR shows the same `report` failure. Not caused by this change; the `acceptance` job itself is green.


obs-gh-matthewgwynne:
I'm fine with this in principal. @obs-gh-abhinavpappu / @obs-gh-sachinjohn - could you review? (I don't have permissions)

client/rest/inbound_share_types_test.go
```diff
@@ -150,3 +150,282 @@ func TestTrackTableResponseMarshaling(t *testing.T) {
...
+	if ref.Id != "" {
+		t.Errorf("expected ref.Id to be empty on decode error, got %q", ref.Id)
+	}
+	if ref.Record != nil {
+		t.Errorf("expected ref.Record to be nil on decode error, got %+v", ref.Record)
+	}
+}
+
+func TestDatasetRef_UnmarshalJSON_AdversarialInputs(t *testing.T) {
+	cases := []struct {
```
obs-gh-abhinavpappu, line 155: nit: could probably do all the above tests as cases in a single test like this one for conciseness/readability

docs/superpowers/specs/2026-05-02-datasetref-migration-design.md
```diff
@@ -0,0 +1,232 @@
+# DatasetRef REST shape migration — Terraform provider design
```
obs-gh-abhinavpappu: i would avoid checking in these 2 files, both because this repo is public, and since these files won't really be relevant in the future; i think the comments already present in the code provide sufficient context

obs-gh-abhinavpappu (approved):

docs/superpowers/specs/2026-05-02-datasetref-migration-design.md
obs-gh-matthewso: Done — removed both files in 3f5c65c3, and replaced the spec §7 reference in DatasetRef's doc comment with an inline description.

client/rest/inbound_share_types_test.go
obs-gh-matthewso: Done — consolidated into a single `TestDatasetRef_UnmarshalJSON` table test with 11 subtests in 3f5c65c3. Every previous assertion is preserved; the zero-on-error invariant is a post-error check on the `RecordWrongType` subtest.

obs-gh-abhinavpappu (approved):

PR #322, opened 2026-05-12
feat: add observe_skill resource and data source

obs-gh-abhinavpappu:
Adds Terraform resource and data source for Observe AI SRE skills using the /v1/skills API (PATCH-on-change, Workspace or Private visibility). Includes acceptance tests, examples, and registry docs consistent with terraform-plugin-docs.


observe/resource_skill.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return diags
+}
+
+func skillCreateRequestFromResourceData(data *schema.ResourceData) *rest.SkillCreateRequest {
+	return &rest.SkillCreateRequest{
+		Label:       data.Get("label").(string),
+		Description: data.Get("description").(string),
+		Content:     data.Get("content").(string),
+		Visibility:  "Listed", // not exposed in TF schema; may be added later
```
vikramraman: Any reason to not expose this? We could set this as an optional field that defaults to Listed?

vikramraman (approved):
lgtm. Left a minor comment on the visibility attribute.

observe/resource_skill.go
obs-gh-abhinavpappu: I was mainly thinking from the perspective of using terraform as a centralized resource management system for the entire org, in which case there's no point in creating Unlisted/Personal skills for some service account user that wouldn't be available for anyone else.

But I guess people could also use terraform individually for their own stuff

obs-gh-abhinavpappu: updated

PR #323, opened 2026-05-18
fix: pin tfplugindocs provider name for git worktrees

obs-gh-abhinavpappu:
tfplugindocs infers the provider name from the checkout directory name, so make generate fails in git worktrees and can corrupt docs. Pass --provider-name terraform-provider-observe so generation matches a standard clone and CI.


obs-gh-abhinavpappu:
included in a different pr and is in master already

PR #324, opened 2026-05-29
feat(monitor_v2): reconcile anomaly schema to typed algorithm fields

obs-gh-enricogiorio:
## Summary

Migrate the anomaly monitor rule template to the typed GraphQL fields the Observe backend now exposes:

- `num_standard_deviations` is nested inside a `basic_algorithm` block and submitted via `basicAlgorithmTyped` rather than the deprecated `basicAlgorithm: JsonObject` placeholder and top-level `numStandardDeviations`.
- A new `seasonal_algorithm` block selects Prophet-based seasonal forecasting, with an optional `sensitivity` tier (`low`, `medium`, `high`, `very_high`).

The two algorithm blocks are mutually exclusive because the backend rejects a monitor configured with both families.

## Test plan

- [ ] `make build` / `go build ./...`
- [ ] `make testacc TESTARGS='-run TestAccObserveMonitorV2Anomaly'` (existing basic-algorithm path)
- [ ] `make testacc TESTARGS='-run TestAccObserveMonitorV2AnomalySeasonal'` (new seasonal path, both with and without `sensitivity`)
- [ ] Confirm an existing TF config that still uses `basic_algorithm` plans clean after upgrade.

### DIFF
--- a/docs/resources/monitor_v2.md
+++ b/docs/resources/monitor_v2.md
@@ -98,10 +98,11 @@ resource "observe_monitor_v2" "anomaly_example" {
 
   rule_template {
     anomaly {
-      value_column_name       = "A_credits_adhoc_query_sum"
-      compare_fn              = "above"
-      num_standard_deviations = 3
-      basic_algorithm {}
+      value_column_name = "A_credits_adhoc_query_sum"
+      compare_fn        = "above"
+      basic_algorithm {
+        num_standard_deviations = 3
+      }
     }
   }
 
@@ -892,12 +893,13 @@ Optional:
 Required:
 
 - `compare_fn` (String) The bound comparison function (Above, Below, AboveOrBelow) defining which direction(s) of standard deviation to consider out of bounds.
-- `num_standard_deviations` (Number) The number of standard deviations a data point must be out of bounds to be marked as anomalous (1 to 5).
 - `value_column_name` (String) Indicates which of the columns in the input query to apply the basic algorithm and create bounds over.
 
 Optional:
 
-- `basic_algorithm` (Block List, Max: 1) Makes the monitor use the basic algorithm. Set to an empty block to enable. The basic algorithm computes the average and standard deviations over the computation window. (see [below for nested schema](#nestedblock--rule_template--anomaly--basic_algorithm))
+- `basic_algorithm` (Block List, Max: 1) Configures the monitor to use the basic standard-deviation anomaly algorithm. Set this block (with `num_standard_deviations`) to enable; the basic algorithm computes the average and standard deviations over the computation window. Mutually exclusive with `seasonal_algorithm`. (see [below for nested schema](#nestedblock--rule_template--anomaly--basic_algorithm))
+- `num_standard_deviations` (Number, Deprecated) The number of standard deviations a data point must be out of bounds to be marked as anomalous (1 to 5). Prefer setting this inside the `basic_algorithm` block; the top-level field is deprecated.
+- `seasonal_algorithm` (Block List, Max: 1) Configures the monitor to use Prophet-based seasonal forecasting. Set this block to enable; data points outside the predicted band are flagged. Mutually exclusive with `basic_algorithm`. (see [below for nested schema](#nestedblock--rule_template--anomaly--seasonal_algorithm))
 
 Read-Only:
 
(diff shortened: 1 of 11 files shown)


observe/resource_monitor_v2.go
```diff
@@ -139,17 +139,38 @@ func resourceMonitorV2() *schema.Resource {
 										DiffSuppressFunc: diffSuppressEnums,
 										Description:      descriptions.Get("monitorv2", "schema", "rule_template", "anomaly", "compare_fn"),
 									},
-									"num_standard_deviations": { // Int64!
```
vikramraman, line 142: This is a breaking change. Do we know whether there are any customers using this via terraform?

We don't have an established policy around this in the provider. But worth checking the impact before merging this. If there is active usage, it might be preferable to mark this field as deprecated and urge users to upgrade first before completely removing this field.

obs-gh-enricogiorio: This should have never been removed in the first place, it was not my intent to do so. I updated the PR to leave the fields intact.

vikramraman: @obs-gh-enricogiorio The PR hasn't been updated? Still shows that field as removed.

docs/resources/monitor_v2.md
```diff
@@ -819,12 +820,20 @@ Optional:
 Required:
 
 - `compare_fn` (String) The bound comparison function (Above, Below, AboveOrBelow) defining which direction(s) of standard deviation to consider out of bounds.
-- `num_standard_deviations` (Number) The number of standard deviations a data point must be out of bounds to be marked as anomalous (1 to 5).
 - `value_column_name` (String) Indicates which of the columns in the input query to apply the basic algorithm and create bounds over.
 
 Optional:
 
+<<<<<<< HEAD
```
vikramraman: Need to fix this, looks to be some merge conflict.

vikramraman (changes requested):

observe/resource_monitor_v2.go
```diff
@@ -1172,15 +1231,27 @@ func monitorV2FlattenRuleTemplate(gqlRuleTemplate gql.MonitorV2RuleTemplate) []i
...
+		"value_column_name": gqlTemplate.ValueColumnName,
+		"compare_fn":        toSnake(string(gqlTemplate.CompareFn)),
 	}
 	if gqlTemplate.ComputationWindow != nil {
 		template["computation_window"] = gqlTemplate.ComputationWindow.String()
 	}
-	if gqlTemplate.BasicAlgorithm != nil {
-		template["basic_algorithm"] = []interface{}{map[string]interface{}{}}
+	if gqlTemplate.BasicAlgorithmTyped != nil {
+		nsdu := int(gqlTemplate.BasicAlgorithmTyped.NumStandardDeviations)
```
obs-gh-abhinavpappu, line 1241: for existing monitors created before `MonitorV2AnomalyRuleTemplate.basicAlgorithmTyped` was introduced, does `MonitorV2AnomalyRuleTemplate.basicAlgorithmTyped.numStandardDeviations` return what `MonitorV2AnomalyRuleTemplate.numStandardDeviations` used to return in the backend?

if not, we might want to still read the top-level field as a fallback to prevent customers from seeing a diff after updating the terraform provider (just a one-time diff so not the worst thing, especially given that there likely aren't *too* many anomaly monitors created yet, but would be nice to avoid if possible)

observe/resource_monitor_v2_test.go
```diff
@@ -1830,3 +1915,94 @@ func TestAccObserveMonitorV2AnomalyWithNoDataRule(t *testing.T) {
 		},
 	})
 }
+
+func TestAccObserveMonitorV2AnomalySeasonal(t *testing.T) {
+	// The CI customer does not yet have the seasonal anomaly capability
+	// enabled, so monitor creation fails with "seasonal anomaly monitors are
+	// not enabled for this customer". Skip until the capability is rolled out
+	// to the test customer.
+	t.Skipf("skipping until seasonal anomaly monitors are enabled for the CI customer")
```
obs-gh-abhinavpappu: please enable this for the CI customer either prior to merge or as a quick follow-up

obs-gh-abhinavpappu (approved):

observe/resource_monitor_v2_test.go
obs-gh-enricogiorio: Done

vikramraman (approved):

PR #329, opened 2026-06-17
Updated GQL schema

georgios-observeinc:
Removed Operations:
- CreateWorkspace
- DeleteWorkspace

Removed Properties:
- Dataset.on_demand_materialization_length

### DIFF
--- a/observe/resource_dataset.go
+++ b/observe/resource_dataset.go
@@ -80,14 +80,6 @@ func resourceDataset() *schema.Resource {
 				Optional:    true,
 				Description: descriptions.Get("dataset", "schema", "path_cost"),
 			},
-			"on_demand_materialization_length": {
-				Type:             schema.TypeString,
-				Optional:         true,
-				Computed:         true,
-				ValidateDiagFunc: validateTimeDuration,
-				DiffSuppressFunc: diffSuppressTimeDuration,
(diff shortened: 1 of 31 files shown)


observe/resource_dataset.go
```diff
@@ -80,14 +80,6 @@ func resourceDataset() *schema.Resource {
 				Optional:    true,
 				Description: descriptions.Get("dataset", "schema", "path_cost"),
 			},
-			"on_demand_materialization_length": {
```
obs-gh-abhinavpappu, line 83: Removing this would be a breaking change and terraform runs for anyone still using `on_demand_materialization_length` would fail at the validation step and block *all* terraform changes.

Options:
1. We acknowledge the breaking change and release a new major version of the provider. Would prefer not to do that.
2. We keep the field but mark it deprecated and make it have no effect anymore. Still technically a breaking change but not as impactful as removing the field from the schema altogether. (And then can eventually remove it altogether along with other deprecated fields when we do release a new major version at some point in the future)
4. We map the field to the layered setting transparently in the terraform provider so that there's effectively no difference from the consumer's perspective.

Would prefer either 2 or 3.

obs-gh-abhinavpappu (changes requested):

georgios-observeinc:
Closing in favor of #332

PR #332, opened 2026-06-18
Updated GQL schema

georgios-observeinc:
Removed Operations:
- CreateWorkspace
- DeleteWorkspace

### DIFF
--- a/observe/resource_dataset.go
+++ b/observe/resource_dataset.go
@@ -85,8 +85,9 @@ func resourceDataset() *schema.Resource {
 				Optional:         true,
 				Computed:         true,
 				ValidateDiagFunc: validateTimeDuration,
-				DiffSuppressFunc: diffSuppressTimeDuration,
+				DiffSuppressFunc: diffSuppressAlways,
 				Description:      descriptions.Get("dataset", "schema", "on_demand_materialization_length"),
+				Deprecated:       "the configuration of on_demand_materialization_length has been migrated to the Dataset.maximumOnDemandMaterializationDays layered setting. Changes to on_demand_materialization_length are no longer respected",
 			},
 			"freshness": {
 				Type:             schema.TypeString,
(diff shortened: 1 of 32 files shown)


observe/data_source_dataset.go
```diff
@@ -75,8 +75,11 @@ func dataSourceDataset() *schema.Resource {
 				Description: descriptions.Get("dataset", "schema", "path_cost"),
 			},
 			"on_demand_materialization_length": {
-				Type:        schema.TypeString,
-				Computed:    true,
+				Type:     schema.TypeString,
+				Optional: true,
+				DiffSuppressFunc: func(string, string, string, *schema.ResourceData) bool {
```
obs-gh-abhinavpappu: it's mainly the resource that needs the diff suppress, not the data source

not sure what exactly diff suppressing the data source would do but i believe it basically has no effect, would remove it from here

observe/resource_dataset.go
obs-gh-abhinavpappu: replace this with `diffSuppressAlways` (from `observe/diff.go`)

obs-gh-abhinavpappu, line 83: add `Deprecated: "<some message about this field no longer being honored; and perhaps a suggestion to use the layered setting instead if we want to point people to that>"`

obs-gh-abhinavpappu (commented):
need to update `resource_dataset_test.go` to no longer assert `on_demand_materialization_length` values in the response

observe/resource_dataset.go
```diff
@@ -85,8 +85,9 @@ func resourceDataset() *schema.Resource {
 				Optional:         true,
 				Computed:         true,
 				ValidateDiagFunc: validateTimeDuration,
-				DiffSuppressFunc: diffSuppressTimeDuration,
+				DiffSuppressFunc: diffSuppressAlways,
 				Description:      descriptions.Get("dataset", "schema", "on_demand_materialization_length"),
+				Deprecated:       "the configuration of on_demand_materialization_length has been migrated to the Dataset.maximumOnDemandMaterializationDays layered setting",
```
obs-gh-abhinavpappu: nit: maybe make it more explicit that changes to `on_demand_materialization_length` are no longer respected

observe/resource_dataset.go
```diff
@@ -85,8 +85,9 @@ func resourceDataset() *schema.Resource {
 				Optional:         true,
 				Computed:         true,
 				ValidateDiagFunc: validateTimeDuration,
-				DiffSuppressFunc: diffSuppressTimeDuration,
+				DiffSuppressFunc: diffSuppressAlways,
 				Description:      descriptions.Get("dataset", "schema", "on_demand_materialization_length"),
```
obs-gh-abhinavpappu, line 89: i thought the generated documentation would automatically be updated when `Deprecated:` was added, but looks like not, could you also update the description for `on_demand_materialization_length` in `observe/descriptions/dataset.yaml`

obs-gh-abhinavpappu (approved):

PR #334, opened 2026-06-25
feat(monitor_v2): support alarm_mode on scheduled monitors

obs-gh-andrewlee:
## Summary

Adds the `alarm_mode` attribute to the cron `scheduling { scheduled { ... } }` block of `observe_monitor_v2`, mapping to the new `MonitorV2CronScheduleInput.alarmMode` GraphQL field (enum `MonitorV2AlarmMode`). Accepts `per_run` (the default) or `ongoing`.

```hcl
scheduling {
  scheduled {
    raw_cron   = "0 0 * * *"
    timezone   = "America/New_York"
    alarm_mode = "ongoing"   # new — defaults to "per_run"
  }
}
```

- `per_run` (default): each evaluation that fires opens its own zero-duration alarm, independent of the previous evaluation.
- `ongoing`: consecutive evaluations re-asserting the same (group, level) extend a single ongoing alarm until an evaluation no longer asserts that level.

## What changed

The enum and the `alarmMode` input field already existed in the synced GraphQL schema / generated client (from #332). This wires it through the provider:

- **`client/internal/meta/operation/monitorv2.graphql`** — request `alarmMode` in the `MonitorV2CronSchedule` fragment so the value round-trips on read.
- **`client/meta/genqlient.generated.go`** — regenerated (`go generate`); adds `AlarmMode` to the read struct (the input struct already had it).
- **`client/meta/helpers.go`** — add `AllMonitorV2AlarmModes` enum slice.
- **`observe/resource_monitor_v2.go`** — schema field (`Optional`, `Default: per_run`, `validateEnums`, `diffSuppressEnums`); expand sets `cron.AlarmMode`; flatten reads it back.
- **`observe/data_source_monitor_v2.go`** — `alarm_mode` as `Computed` (shares the resource flatten).
- **`observe/descriptions/monitorv2.yaml`** + **`docs/resources/monitor_v2.md`** + **`docs/data-sources/monitor_v2.md`** — documentation.
- **`observe/resource_monitor_v2_test.go`** — extends `TestAccObserveMonitorRawCron` with an explicit `ongoing` case plus a `per_run`-default round-trip step (guards against perpetual diff).

The field follows the existing `operator` enum precedent (#248), with `diffSuppressEnums` added (matching the `rule_kind` pattern). Using `Default: per_run` matches the server default and keeps existing cron monitors diff-free.

## Testing

- `go build ./...`, `go vet ./observe/... ./client/meta/...`, `gofmt` — all clean.
- `TestAccObserveMonitorRawCron` compiles and skips without a backend. Run with a live backend: `TF_ACC=1 go test ./observe/ -run TestAccObserveMonitorRawCron -v`.

### DIFF
--- a/client/internal/meta/operation/monitorv2.graphql
+++ b/client/internal/meta/operation/monitorv2.graphql
@@ -184,6 +184,7 @@ fragment MonitorV2TransformSchedule on MonitorV2TransformSchedule {
 fragment MonitorV2CronSchedule on MonitorV2CronSchedule {
     rawCron
     timezone
+    alarmMode
 }
 
 fragment MonitorV2Scheduling on MonitorV2Scheduling {
--- a/client/meta/genqlient.generated.go
+++ b/client/meta/genqlient.generated.go
@@ -7614,6 +7614,9 @@ type MonitorV2CronSchedule struct {
 	// A timezone is required to ensure that interpretation of scheduling on the wall-clock
 	// is done relative to the desired timezone.
 	Timezone string `json:"timezone"`
+	// Controls how alarms are emitted across consecutive monitor evaluations.
+	// Defaults to PerRun when unset.
+	AlarmMode *MonitorV2AlarmMode `json:"alarmMode"`
 }
 
 // GetRawCron returns MonitorV2CronSchedule.RawCron, and is useful for accessing the field via an interface.
@@ -7621,6 +7624,9 @@ func (v *MonitorV2CronSchedule) GetRawCron() *string { return v.RawCron }
 
 // GetTimezone returns MonitorV2CronSchedule.Timezone, and is useful for accessing the field via an interface.
 func (v *MonitorV2CronSchedule) GetTimezone() string { return v.Timezone }
+
+// GetAlarmMode returns MonitorV2CronSchedule.AlarmMode, and is useful for accessing the field via an interface.
+func (v *MonitorV2CronSchedule) GetAlarmMode() *MonitorV2AlarmMode { return v.AlarmMode }
 
 type MonitorV2CronScheduleInput struct {
 	RawCron   *string             `json:"rawCron"`
@@ -17493,6 +17499,7 @@ fragment MonitorV2TransformSchedule on MonitorV2TransformSchedule {
 fragment MonitorV2CronSchedule on MonitorV2CronSchedule {
 	rawCron
 	timezone
+	alarmMode
 }
 fragment MonitorV2ComparisonTerm on MonitorV2ComparisonTerm {
 	comparison {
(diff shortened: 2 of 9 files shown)


obs-gh-abhinavpappu (commented):
It appears that the API rejects requests where `alarmMode` is non-nil for tenants that do not have the feature flag enabled. And SlowLane tenants don't have the flag enabled yet, so we want to make sure we don't *explicitly* send any value for `alarmMode` when `alarm_mode` is not set in the terraform config.

Adding `Default: toSnake(string(gql.MonitorV2AlarmModePerrun))` will result in the terraform provider explicitly sending `alarmMode: per_run` in the request when `alarm_mode` is not set.

I would suggest one of the following:
- If the API returns `null` in the response for pre-existing monitors that don't have an explicit alarmMode set (which seems to be the case):
  - Remove the `Default`
  - Instead of `data.Get(fmt.Sprintf("%salarm_mode", path))`, use `data.GetOk` and only set `cron.AlarmMode` if `ok == true`
- If the API returns `per_run` in the response for pre-existing monitors:
  - Remove the `Default`
  - Add `Computed: true` to prevent diffs
    - this is in general the better pattern for backend defaults, and also allows the default to change in the backend without the terraform provider pinning it to a certain value for unset fields
  - Use `GetRawConfig` to get the value for `alarm_mode`. See https://github.com/observeinc/terraform-provider-observe/blob/6b95650f5aaa20f439fbff3cc35846536172b974/observe/resource_reference_table.go#L145 for an explanation of why that instead of `GetOk`



> Using Default: per_run matches the server default and keeps existing cron monitors diff-free.

Not sure I'm fully following the reasoning for `Default` here based on the behavior I'm observing, but let me know if I'm missing something!

obs-gh-andrewlee:
@obs-gh-abhinavpappu thanks, you're right — pushed a fix (`aee95ba3`).

I confirmed the backend behavior in the monorepo before deciding:
- `resolveMonitorV2FeatureFlags` (`gql_monitorv2.go`) rejects **any** non-nil `alarmMode` when `FlagEnableOngoingAlarms` is off — not just `Ongoing` — so the old `Default: per_run` really would 500 create/update on flagless tenants.
- The stored `MonitorV2CronSchedule.alarmMode` is nullable and **returns nil for unset** (no server-side default on the `scheduling` read path), and `UpdateFromInput` is a full replace.

Since that's the null-read case, I went with your first suggestion:
- removed `Default`
- left it `Optional` (no `Computed`)
- the expander now gates on `GetOk` and only sets `cron.AlarmMode` when the user explicitly set `alarm_mode` (mirrors the sibling `raw_cron`)

So an unset `alarm_mode` now sends no `alarmMode` at all and round-trips cleanly with no diff.

Tests added:
- unit tests (no backend / no flag) asserting the expander returns `nil` when unset, plus flatten round-trip and the schema flags (guards against a stray `Default`/`Computed` regression)
- an acceptance test covering the lifecycle (unset → ongoing → case-insensitive no-drift → per_run → clear) and an invalid-value `ExpectError`

Also reworded the doc/description so "`per_run`" is described as the backend's default-when-unset rather than implying the provider writes it. The value-setting acceptance steps require `FlagEnableOngoingAlarms` on the test tenant; the unset/invalid paths don't.

obs-gh-abhinavpappu (approved):

PR #341, opened 2026-07-13
fix(build): restore ../observe as OBSERVE_ROOT default

obs-gh-sachinjohn:
The ~/observe value was introduced incidentally in an unrelated sed-fix commit and never expands: Make does not expand ~ and quoted recipe usages suppress shell tilde expansion. Revert to the sibling ../observe convention


obs-gh-peterkolloch:
Not sure what the CI failure is about, LGTM to me.

"Discussion" here: https://snowflake.slack.com/archives/C0AD3EE0NSJ/p1783846717523089

obs-gh-abhinavpappu (approved):

PR #344, opened 2026-07-29
feat: warn on `terraform apply` for inefficient acceleration.

sfc-gh-ojiang:
Adds acceleration_type as a computed read-only field to the observe_dataset resource and data source, and emits a Terraform warning on apply when a dataset's pipeline results in acceleration type other.

Dataset acceleration falls into four types:
- insert_only (most efficient)
- aggregation (efficient)
- other (likely expensive)
- not_supported

Unlike datasets created in the UI, datasets created in terraform do not receive this AccelerationType information, making it easy to accidentally create expensive datasets without feedback. This change adds a warning on `terraform apply` if acceleration falls under the 'other' type. The 'not_supported' type is already being caught as an error.

Changes
- Adds accelerationType to the Dataset GraphQL fragment and regenerates the client
- Exposes acceleration_type as a computed field on the resource and data source
- Emits a warning diagnostic on create and update when acceleration_type == "other"

Notes
- Warning fires after apply, not during plan. terraform-plugin-sdk/v2's CustomizeDiff only supports returning errors — there is no mechanism to surface a warning at plantime. Showing this before the "yes" prompt would require migrating to terraform-plugin-framework, which supports diag.Diagnostics in PlanModifiers.
- acceleration_type is (known after apply) during plan. Seems to be because acceleration_type is not returned in preflight mode, so surfacing acceleration_type during plan will require backend change.
- I opted not to use PipelineWarning.ExpensiveAcceleration (which is more definitive than acceleration_type==other which is only "likely expensive"). The checkQueries GraphQL operation compiles a pipeline without saving it and returns PipelineWarning results, including an ExpensiveAcceleration kind. However, checkQueries is not currently wired up in this provider and would require a new GraphQL operation and client method. 

Example terraform output:
[demo.txt](https://github.com/user-attachments/files/30484884/demo.txt)

### DIFF
--- a/client/internal/meta/operation/dataset.graphql
+++ b/client/internal/meta/operation/dataset.graphql
@@ -7,6 +7,7 @@ fragment Dataset on Dataset {
 	iconUrl
 	accelerationDisabled
 	accelerationDisabledSource
+	accelerationType
 	version
 	lastSaved
 	pathCost
--- a/client/meta/genqlient.generated.go
+++ b/client/meta/genqlient.generated.go
@@ -19,6 +19,25 @@ const (
 	AccelerationDisabledSourceView    AccelerationDisabledSource = "View"
 )
 
+type AccelerationType string
+
+const (
+	// Dataset can be accelerated by inserting new rows as they come in. This is
+	// the most efficient form of acceleration.
+	AccelerationTypeInsertonly AccelerationType = "InsertOnly"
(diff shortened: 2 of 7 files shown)


obs-gh-abhinavpappu (approved):
lgtm!

have been wanting to migrate to the terraform framework which would let us emit warnings on plan, but haven't had the time yet

PR #347, opened 2026-07-31
feat(dataset): surface downstream compile breaks as Terraform warnings

sfc-gh-abpappu:
## Summary

When a dataset save breaks a downstream dataset's compilation, or when a dataset itself has a compilation error, Terraform now surfaces these as warnings rather than silently discarding them.

Previously, the provider called `DatasetSaveResult` but threw away the `errorDatasets` list, and `GetDataset`/`GetLogDerivedMetricDataset` never requested `compilationError`. This meant broken downstream datasets were invisible to `terraform apply` and `terraform plan` (read/refresh path).

### GQL fragments (`dataset.graphql`, `log_derived_metrics.graphql`)
Added `compilationError { error errorInDatasetId }` to the `Dataset` and `LogDerivedMetricDataset` fragments so `GetDataset` returns compilation state.

### `SaveDataset` / `SaveLogDerivedMetricDataset` return type change
Both methods now return the full `*DatasetSaveResult` / `*LogDerivedMetricDatasetSaveResult` instead of just the inner `*Dataset`. This exposes `ErrorDatasets` to callers without requiring a separate method.

### Apply-time warnings (Create / Update)
After a successful save, `appendDownstreamWarnings` emits a `diag.Warning` for each entry in `ErrorDatasets` where `HasExistingError == false` — i.e., compile breaks this change newly introduced, not ones that were already broken before.

### Read-time warnings (Read / plan refresh)
After `GetDataset`/`GetLogDerivedMetricDataset`, `appendCompilationWarning` / `appendLDMCompilationWarning` emits a warning if `CompilationError` is non-nil, distinguishing between an error in the dataset's own OPAL vs. one inherited from an upstream dependency.

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -72,7 +72,7 @@ func (c *Client) GetDataset(ctx context.Context, id string) (*meta.Dataset, erro
 	return c.Meta.GetDataset(ctx, id)
 }
 
-func (c *Client) SaveDataset(ctx context.Context, wsid string, input *meta.DatasetInput, queryInput *meta.MultiStageQueryInput, dependencyHandling *meta.DependencyHandlingInput) (*meta.Dataset, error) {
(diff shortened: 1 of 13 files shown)


obs-gh-christiannester:
How far it this change? Overall this looks good to me.

PR #357, opened 2026-08-07
Adopt existing correlation tags instead of failing on create

sfc-gh-abpappu:
## Summary

A correlation tag has no ID of its own -- it's identified entirely by its `(dataset, tag, column, path)` combination, not just by dataset and name: the same tag name can be attached to multiple columns (or paths) on the same dataset, each a distinct mapping. A matching tag that already exists under that full combination is already in the desired state. `observe_correlation_tag` create now checks whether that exact combination already exists and, if so, adopts it (a no-op) instead of failing -- for example when it was previously created through the UI, or by another `apply`.

## Testing

Added `TestCorrelationTagAdoptExisting`, which creates the tag out-of-band via the API client before applying the `observe_correlation_tag` resource, and asserts the apply succeeds with no error and no subsequent plan diff. It also applies a second resource sharing the same `(dataset, name)` but a different column, and asserts it's created as its own distinct mapping rather than being adopted into the first.

Tags created by system content packages aren't special-cased here -- the backend doesn't expose tag creator info over GraphQL today, so adopting one behaves like adopting any other tag. If it's later destroyed, the backend's existing restriction on deleting system-created tags still applies unchanged.

### DIFF
--- a/docs/resources/correlation_tag.md
+++ b/docs/resources/correlation_tag.md
@@ -4,10 +4,19 @@ page_title: "observe_correlation_tag Resource - terraform-provider-observe"
 subcategory: ""
 description: |-
   A correlation tag can be attached to columns of a dataset. These tags are later used to correlate multiple datasets.
+  A correlation tag is uniquely identified by its (dataset, name, column, path)
+  combination rather than by a separate ID. If a correlation tag with the same
+  combination already exists -- for example, created previously through the UI --
+  this resource adopts it instead of failing.
 ---
 # observe_correlation_tag
 
 A correlation tag can be attached to columns of a dataset. These tags are later used to correlate multiple datasets.
+
+A correlation tag is uniquely identified by its (dataset, name, column, path)
+combination rather than by a separate ID. If a correlation tag with the same
+combination already exists -- for example, created previously through the UI --
+this resource adopts it instead of failing.
 ## Example Usage
 ```terraform
 data "observe_workspace" "default" {
--- a/observe/resource_correlation_tag.go
+++ b/observe/resource_correlation_tag.go
@@ -101,13 +101,28 @@ func resourceCorrelationTagCreate(ctx context.Context, data *schema.ResourceData
 		return diags
 	}
 
-	err := client.CreateCorrelationTag(ctx, params.Dataset, params.Tag, params.Path)
+	// A correlation tag isn't a first-class object with its own identity -- it's
+	// just a (dataset, tag, column, path) combination, and that combination is
+	// its entire state. If it already exists (e.g. someone created it in the UI,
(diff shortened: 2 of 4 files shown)


docs/resources/correlation_tag.md
```diff
@@ -4,10 +4,19 @@ page_title: "observe_correlation_tag Resource - terraform-provider-observe"
 subcategory: ""
 description: |-
   A correlation tag can be attached to columns of a dataset. These tags are later used to correlate multiple datasets.
+  A correlation tag is uniquely identified by its (dataset, name, column, path)
```
obs-gh-christiannester, line 7: This is a quadruple while the text below and the PR takes about a triple. Can we make that consistent?
Does it make sense to include the path into the identification? Then two tags on the same dataset with the same name but different paths would be valid.

observe/resource_correlation_tag.go
```diff
@@ -101,15 +101,30 @@ func resourceCorrelationTagCreate(ctx context.Context, data *schema.ResourceData
 		return diags
 	}
 
-	err := client.CreateCorrelationTag(ctx, params.Dataset, params.Tag, params.Path)
+	// A correlation tag isn't a first-class object with its own identity -- it's
+	// just a (dataset, tag, path) triple, and that triple is its entire state. If
```
obs-gh-christiannester: Here is says triple.

observe/resource_correlation_tag_test.go
```diff
@@ -52,3 +56,63 @@ func TestCorrelationTagCreation(t *testing.T) {
 		},
 	})
 }
+
+// TestCorrelationTagAdoptExisting verifies that creating a correlation tag that
```
obs-gh-christiannester, line 60: How should the case when we have a new TF tag with the same dataset and the same tag name but a different path be handled? Should it update the existing tag or create a second one with the same name? Could you add the corresponding test case as well?

obs-gh-christiannester (changes requested):

docs/resources/correlation_tag.md
obs-gh-abhinavpappu: Updated. And yes it is indeed the case that you can how 2 tags on the same dataset with the same name but on different columns & paths. Tags do truly seem to be identified by the full quadruple and not just the dataset + name.

observe/resource_correlation_tag_test.go
obs-gh-abhinavpappu: It creates a second one with the same name, added a test

sfc-gh-cnester (approved):

obs-gh-christiannester:
Looks good to me

obs-gh-christiannester (approved):

PR #358, opened 2026-08-10
fix(query): stop perpetual plan diff from inputs no stage uses

sfc-gh-abpappu:
## Problem

An input declared in `inputs` that no stage uses — neither named as a stage's `input` nor referenced as `@name` in a pipeline — is dropped when the query is written. `appendReferencedInputs` attaches inputs to individual stages and `MultiStageQueryInput` has no top-level input list, so the input never reaches the API. It is then absent from the read while the configuration still declares it, so every plan shows it as a pending addition that applying never resolves.

Affects `observe_dataset`, `observe_monitor`, and `observe_monitor_v2`, which share `flattenAndSetQuery`.

## Fix

`flattenAndSetQuery` carries such an input forward from the prior value rather than letting it fall out of state, and warns that the declaration has no effect. It now returns `diag.Diagnostics` so the warning reaches the caller.

Two cases are deliberately left alone. Distinguishing either one requires the API response, which is why this lives on the read path rather than in a plan-time or write-path check:

- An input a stage does reference but that is missing from the response is a real discrepancy, so it is not carried forward and still surfaces as a diff.
- An input the API does return is never warned about. This covers inputs Observe derives from `^`"link"` traversals; `inputIsUsed` does not count a traversal as a use, since the traversal names the link rather than the input.

Existing state converges after one apply. The first plan still shows the addition, applying is a no-op against Observe, and the read that follows carries the input forward.

## Testing

`TestFlattenAndSetQueryUnusedInputs` in `observe/resource_dataset_test.go` covers the carry-forward for inputs no stage uses, and verifies that a used input removed outside Terraform still surfaces as a diff.

### DIFF
--- a/observe/resource_dataset.go
+++ b/observe/resource_dataset.go
@@ -412,10 +412,8 @@ func datasetToResourceData(d *gql.Dataset, data *schema.ResourceData) (diags dia
 	}
 
 	if d.Transform != nil && d.Transform.Current != nil && d.Transform.Current.Query != nil {
-		_, err := flattenAndSetQuery(data, d.Transform.Current.Query.Stages, d.Transform.Current.Query.OutputStage)
-		if err != nil {
-			diags = append(diags, diag.FromErr(err)...)
-		}
+		_, queryDiags := flattenAndSetQuery(data, d.Transform.Current.Query.Stages, d.Transform.Current.Query.OutputStage)
+		diags = append(diags, queryDiags...)
 	}
 
 	if err := data.Set("oid", d.Oid().String()); err != nil {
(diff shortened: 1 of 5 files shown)


observe/resource_dataset.go
```diff
@@ -456,8 +456,32 @@ func flattenAndSetQuery(data *schema.ResourceData, gqlstages []gql.StageQuery, o
...
+			Severity: diag.Warning,
+			Summary:  fmt.Sprintf("input %q is not used by any stage", name),
+			Detail: "Observe does not save inputs that no stage references, so this one is not stored. " +
+				"Remove it from inputs, or reference it from a stage pipeline as @" + name + ".",
+		})
+	}
+
 	if err := data.Set("inputs", inputs); err != nil {
-		return nil, err
+		return nil, diag.FromErr(err)
```
obs-gh-christiannester: Below you are doing an append to diags before returning the error. Here you loose the possible warnings.

obs-gh-christiannester (approved):
I am good with this change but it's not clear to me why we change it this way.
Shouldn't an unreferenced stage be an error? We could have started with warning so that people can clean up their configs but why do we carry the stage forward and hide the problem instead of targeting an error?
Or is that a first step for a better handling with warning and then later we switch to an error? If this is the plan the warning should be more explicit that this is going to happen.

Could you also have a look at the PR description. It reads very AI generated. I find it very hard to understand the problem from it although the problem isn't that difficult.

sfc-gh-abpappu:
I guess I was overall thinking from the perspective of preventing the diffs while minimizing disruption.

Ideally we should have had this scenario error when datasets were originally introduced, but since that's not the case, I can't introduce the error now as it could break existing terraform configs (which could potentially have hundreds of offending datasets that need to be fixed to unblock terraform runs), plus some customers have heavy automation surrounding terraform and such errors would break their pipeline and prevent all changes.

My original plan was to just leave the warning in perpetuity and customers could address it to prevent the continuous nagging on every plan. But on second thought, I think switching to erroring in the future makes sense, updated the message. Would also be helpful if we could get telemetry surrounding these warnings so that we can see when they get addressed, will think about that a bit more

observe/resource_dataset.go
sfc-gh-abpappu: done

PR #359, opened 2026-08-10
fix: prevent perpetual diff when backend prunes unreachable pipeline stages

sfc-gh-abpappu:
## Summary

When an `observe_dataset` config declares a pipeline stage whose output nothing consumes, the Observe backend silently prunes it after save. On the next `terraform plan`, the provider reads back N−1 stages while the config still declares N, producing a perpetual diff that never converges — applying doesn't help because the backend prunes the stage again on every save.

The fix re-inserts pruned orphan stages back into Terraform state during the read path (`flattenAndSetQuery`), keeping state in sync with config so the diff disappears. Orphan stages continue to flow to the backend on every apply; the fix only changes what the provider writes to state, not what it sends to the API.

## Details

- **`stagesFromList` and `detectOrphanStages`** added to `data_source_query.go`. `detectOrphanStages` does a backward BFS from the output stage, correctly handling `output_stage: true` on non-last stages and `@alias` references in pipeline text.
- **`flattenAndSetQuery`** in `resource_dataset.go` calls `detectOrphanStages` against the config stage list and re-inserts identified orphans at their original positions before writing state. A safety check (`len(orphanSet) == len(configRaw)-len(stages)`) ensures the merge only runs when the entire count difference is explained by detected orphans, so unrelated backend divergences are never masked.
- Covers `observe_dataset`, `observe_monitor`, and `observe_monitor_v2` since `flattenAndSetQuery` is shared by all three.

## Testing

11 unit tests for `detectOrphanStages` in `observe/data_source_query_unit_test.go` covering: no-orphan forward chains, explicit-input bypass, `@alias` pipeline references, `output_stage` mid-list with a later dependency, two consecutive orphans, anonymous orphan, and a middle orphan in a 3-stage chain.


observe/data_source_query.go
```diff
@@ -248,12 +333,22 @@ func newQuery(data ResourceReader) (*gql.MultiStageQueryInput, diag.Diagnostics)
 		stages = append(stages, stage)
 	}
 
+	var diags diag.Diagnostics
+	for _, idx := range detectOrphanStages(stages) {
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Warning,
+			Summary:  "Stage has no effect",
+			Detail:   fmt.Sprintf("stage %d's output is not used by any subsequent stage, so the Observe backend will not persist it. Terraform will keep showing a diff for this stage on every plan. Remove it, or make a later stage read from it explicitly.", idx),
```
obs-gh-christiannester: Should we inform people in the warning that we plan to turn this into an error?

obs-gh-christiannester (approved):

observe/data_source_query.go
sfc-gh-abpappu: updated

PR #361, opened 2026-08-11
fix: return null object_tags when no tags are set

sfc-gh-abpappu:
## Summary

When the API returns no object tags, `setObjectTagsFromAPI` now writes `nil` to state instead of an empty map. This lets the export flow correctly omit the field — previously, the empty map caused `object_tags = {}` to appear in exported configs even for resources with no tags set.


obs-gh-abhinavpappu:
never mind, i misunderstood the issue i was trying to fix

